Issue #80, opened 2015-03-14
ValueError: too many values to unpack


GoogleCodeExporter:
```
搜索value后程序无响应，shell显示
Traceback (most recent call last):
  File "/usr/share/gameconqueror/GameConqueror.py", line 936, in data_worker
    addr, cur_value, scanmem_type, valid = row
ValueError: too many values to unpack
```

Original issue reported on code.google.com by `china.fr...@gmail.com` on 25 Nov 2014 at 12:27

sriemer:
Sorry, this is my fault. Commit 94c3b9545a5dda1f9d2828f84541eba27fbce935 "gui: fix a hang when searching in scan results" fixes this.

sriemer:
The fix has been merged, hence closing.

Issue #77, opened 2015-03-14
GUI won't start


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. compile and install scanmem & GameConqueror from svn or git
2. run GameConqueror
3. enjoy the error message (if you have Python 3 as default)

What is the expected output? What do you see instead?
Expected: GUI
What I get:
** (GameConqueror.py:20621): WARNING **: Can't load fallback CSS resource: 
Failed to import: The resource at '/org/gnome/adwaita/gtk-fallback.css' does 
not exist

** (GameConqueror.py:20621): WARNING **: Can't load fallback CSS resource: 
Failed to import: The resource at '/org/gnome/adwaita/gtk-fallback.css' does 
not exist
Traceback (most recent call last):
  File "/usr/share/gameconqueror/GameConqueror.py", line 1029, in <module>
    GameConqueror().main()
  File "/usr/share/gameconqueror/GameConqueror.py", line 318, in __init__
    self.backend = GameConquerorBackend()
  File "/usr/share/gameconqueror/backend.py", line 43, in __init__
    self.send_command('reset')
  File "/usr/share/gameconqueror/backend.py", line 67, in send_command
    self.lib.backend_exec_cmd(ctypes.c_char_p(cmd))
TypeError: bytes or integer address expected instead of str instance


What version of the product are you using? On what operating system?
I have tried both svn and git versions. I'm running Manjaro Linux (x64 
OpenBox). Python 3 is the default, which seems to be the problem.

Please provide any additional information below.
Runs fine if using Python 2 (python2 command instead of python).

```

Original issue reported on code.google.com by `valeri.g...@gmail.com` on 23 Aug 2014 at 11:24

sriemer:
I can confirm that with python 3.4. I'm looking for a solution. Ideas welcome as I'm not best in python.

sriemer:
The fixes have been merged, hence closing.

Issue #73, opened 2015-03-14
KeyError: 'I16 I8'


GoogleCodeExporter:
```
Tried to find some values with 'int'

  self.progress_watcher_id = GObject.idle_add(self.progress_watcher)
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
...............................info: we currently have 54576 matches.
/usr/share/gameconqueror/GameConqueror.py:878: PyGIDeprecationWarning: 
source_remove is deprecated; use GLib.source_remove instead
  GObject.source_remove(self.progress_watcher_id)
...........info: we currently have 54420 matches.
...........info: we currently have 54420 matches.
...........info: we currently have 54420 matches.
...........info: we currently have 6 matches.
Traceback (most recent call last):
  File "/usr/share/gameconqueror/GameConqueror.py", line 934, in data_worker
    new_value = self.read_value(addr, TYPENAMES_S2G[scanmem_type.strip()], cur_value)
KeyError: 'I16 I8'

```

Original issue reported on code.google.com by `i.gnaten...@gmail.com` on 16 Jun 2014 at 5:03

sriemer:
Looks like the same issue I've encountered. Commit 8df1810b17f59d4de2ed1e1399b1908394f39961 "gui: use split() instead of strip() in data_worker" should fix it.

sriemer:
The fix has been merged, hence closing.

PR #85, opened 2015-03-21
Fix GameConqueror hanging and Python 3 support

sriemer:
The first commit is the most important one. My own commit aedf069 ("gui: add match offset and region type support") didn't catch all code locations to be modified so that GC is hanging when searching in scan results. Sorry for that!
All other fixes are Python 3 fixes. I've tested them against Python 3.4 and 2.7. I've used the int32 lives value of the HeroAircraft object of the game Chromium B.S.U. to test this. I've frozen the lives to 9 to check if GC does the right thing.

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -936,7 +936,7 @@ class GameConqueror():
                 row = self.scanresult_liststore[i]
                 addr, cur_value, scanmem_type, valid, off, rtype = row
                 if valid:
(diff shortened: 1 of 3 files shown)


sriemer:
This fixes the issues #80 "ValueError: too many values to unpack", #77 "GUI won't start" and #73 "KeyError: 'I16 I8'". So please help reviewing and testing this.

coolwanglu:
LGTM. Thanks!
I don't have a Linux with X now, let me see if I can find one and test it later.

PR #86, opened 2015-03-22
debian, autogen: fix the Debian package build with sbuild

sriemer:
Using sbuild to build Debian packages in a chroot environment is quite common. I'm using a Debian Wheezy chroot environment and noticed several issues. This patch set fixes them all.


sriemer:
Btw.: The co-author of the 'wrap-and-sort' script Benjamin Drung works for us at ProfitBricks at the moment.

coolwanglu:
Thanks!

Issue #88, opened 2015-03-31
Commit 944a54a0 ("Configurable endianness of data") breaks the build


sriemer:
Commit 944a54a0 ("Configurable endianness of data") breaks the build even on Debian Wheezy on powerpc. I've tested the standard Debian Wheezy QEMU VM from https://people.debian.org/~aurel32/qemu/powerpc/. People should test their stuff properly before sending out pull requests,...

````
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -O2 -g -Wall -g -O2 -c endianness.c  -fPIC -DPIC -o .libs/endianness.o
In file included from endianness.c:28:0:
endianness.h:30:1: error: initializer element is not constant
````

AC_C_BIGENDIAN is a much better detection. I'm implementing this at the moment to fix this mess.

coolwanglu:
Closed via #89.

PR #89, opened 2015-03-31
fix endianness check

sriemer:
Commit 944a54a0 ("Configurable endianness of data") breaks the build on some systems due to non-standard endianness detection. So replace that with AC_C_BIGENDIAN added to configure.ac and a check if this defines WORDS_BIGENDIAN.

Tested on x86, x86_64 and in a Debian Wheezy ppc QEMU VM by adding "#warning big endian detected" to the "big_endian = true" case.

Fixes: Github issue #88


coolwanglu:
LGTM. Thanks!

PR #93, opened 2015-04-11
Libless, bypass library dependency

Funmungus:
Enable the configure option --enable-libless.  Using this option will use libreplacements.h and libreplacements.c instead of linking with libraries.  The functions provided have less features, but are more portable.


Funmungus:
Hello.  I thought this option might be helpful for systems that are difficult to get readline support.  Too presumptuous?

sriemer:
Hi,

the idea is good. Especially also when cross compiling it is required to replace libreadline with something else. But I'd like to delay this until scanmem 0.15 is tested and released properly before breaking it again with new features.

You also really need to rework your commits. These aren't ready for a review, yet. Write atomic commits, don't put merge commits in between (use "git fetch upstream" and "git rebase upstream/master" instead), write short descriptions in an imperative mood as if you were giving orders to the code base to change its behavior. Use a prefix as the code base is big. With that it is easier to catch on the first sight which files are touched. This should be a patch set. So the functionality should grow step by step commit by commit but everything has to be compilable after each commit and testable separately. Otherwise there is no chance to review and test big patch sets in very limited time. Fixes of still unreleased commits have to be merged into the original commits. Don't mix unrelated commits like .gitignore changes in between the patch set. Put these at the end of the patch set. So please clean up your mess before sending pull requests for review. Send another pull request once you have reworked your commits into reviewable portions of code. If you need to have a look at how very good commits look like, you could have a look at the Linux kernel source code commits and the file https://www.kernel.org/doc/Documentation/SubmittingPatches.

I hope that this helps.

sriemer:
Btw: GitHub says "We can’t automatically merge this pull request." so this isn't mergable anyway.

Funmungus:
Thank you for the help and suggestions.  I will do this task again after 15 is released.

I have not used github before, only private servers, so I do appreciate the insight.

Funmungus:
In that case I propose closing this until 15 is released.  I will not lose my place.

sriemer:
Thanks for understanding. You can also send further PRs in between so that we can review them before. But commits must be reviewable.

PR #96, opened 2015-04-16
Libless

Funmungus:
Enable bypassing library linking with --enable-libless option in configure script.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -31,7 +31,12 @@ libscanmem_la_SOURCES = commands.h \
     target_memory_info_array.h \
     target_memory_info_array.c \
     value.h \
-    value.c 
+    value.c
+
+if ENABLE_LIBLESS
+  libscanmem_la_SOURCES += libreplacements.h \
+      libreplacements.c
+endif 
 
 libscanmem_la_LDFLAGS = -version-info 1:0:0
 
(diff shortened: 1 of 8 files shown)


sriemer:
Thanks for implementing my hints! This time I could catch on the first sight what your patch set is about to do and this time it is mergable.

While looking at your code I've noticed a couple of things:
* we need your real name in copyrights, commit author and all other places
* coding style isn't correct yet
* it is a good idea to split inclusion of libreplacements from the automake change to build these files in the first commit as different source areas are touched (code and build)
* short description of the third patch is too long (use 70-75 chars at max.)
* long descriptions would be nice

Your real name is required as you're contributing to a public FOSS project. We have to maintain it in the end and in case you've stolen this work from somebody else or are not allowed to publish it, we need to know who to sue. It's also better for you as you can clearly claim that these commits are your work and not from somebody else. There is no way to enforce your copyright otherwise.

In terms of coding style: Look around. Which style has the existing code? Indention is done with four spaces and not only two. Don't put any spaces into "#include". It's done nowhere else this way. Keeping it together makes it easy to grep for includes. Declaring variables in the middle of a function or {} block is bad style and doesn't conform to all C standards.

I'll continue with the technical review and testing your code.

Funmungus:
My pleasure.  This is my first public project, so I am still getting used to things.
I will try to roll back and do your suggestions.

Funmungus:
Where do you mean for me to put long descriptions? In commit messages?

Funmungus:
I sent another PR with what I think you were asking for.  Same branch name.

PR #97, opened 2015-04-18
Libless

Funmungus:
Enable bypassing library linking with --enable-libless option in configure script.  This will allow building without libraries, but less features will be available.
libreplacements implements missing functions.

To build, `./configure --prefix=/usr --enable-libless && make`

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,6 +33,11 @@ libscanmem_la_SOURCES = commands.h \
     value.h \
     value.c 
 
+if ENABLE_LIBLESS
+  libscanmem_la_SOURCES += libreplacements.h \
+      libreplacements.c
+endif
+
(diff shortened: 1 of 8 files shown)


Funmungus:
I just noticed the variables you mentioned, declared at assignment time in readline.  I hope you will accept this merge, with the last commit fixing that code style issue of mine.

sriemer:
There are more problems. I'll fix them for you and send you the final patch set via email so that you can send the final PR.

Funmungus:
The strange readline functions were probably from online searches for the exact definitions.  #if defined also allows for other logical statements, like != 0, which was not available to me with #ifdef.  I do not remember what I put for the include guard, but that should have been an #ifndef.

Funmungus:
I did not know git am complains about empty final line.  I will try to keep that in mind.

PR #98, opened 2015-04-21
Libless

Funmungus:
Enable bypassing library linking with --enable-libless option in configure script. This will allow building without libraries, but less features will be available.
libreplacements implements missing functions.

To build, ./configure --prefix=/usr --enable-libless && make

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -36,16 +36,6 @@ AC_TYPE_SSIZE_T
 AC_C_BIGENDIAN
 
 
-# check if termcap is present, sometimes required by readline
-AC_CHECK_LIB([termcap], [tgetent], [], [])
-
-# check for libreadline
-AC_CHECK_LIB([readline], [readline], [], [
-    echo "libreadline could not be found, which is required to continue."
-    echo "The libreadline-dev package may be required."
-    exit 1
-])
-
 # also need to check if the file is zero'ed (some hardened systems)
 AC_CHECK_FILE([/proc/self/maps], [], [
     echo "This system does not seem to have /proc/pid/maps files."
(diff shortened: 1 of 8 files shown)


Funmungus:
So I learned how to use `git am` today.  Thank you @sriemer

sriemer:
LGTM. Thanks!

menu.c
```diff
@@ -32,8 +32,12 @@
 #include <string.h>
 #include <stdbool.h>
 
+#if defined(HAVE_LIBREADLINE) && HAVE_LIBREADLINE
```
coolwanglu: why not just `#if HAVE_LIBREADLINE` here

coolwanglu:
`enable-libless` looks confusing to me. In fact, the moment I saw this option, I went googling `libless`, thinking it is a new library, until I came back. Not sure if it's only me.

From my appearance, usually this kind of switch is named `--with-libreadline` or `--without-libreadline`. Since we have only two libraries here, it should not be hard.

menu.c
sriemer: Would be also okay, yes. If it would be not defined, then configure would complain.

sriemer:
You're right. It was also confusing for me but I thounght it would be a rare case and I've found so many problems in this feature that I had no better way in mind at that time. Would you like to provide a fix commit?

Funmungus:
I went with --enable because it was available in the autotools. I just have a lack of autotools knowledge, so I am in no way attached to the names I gave things.

configure.ac
```diff
@@ -134,5 +124,27 @@ AC_ARG_ENABLE(gui, [AS_HELP_STRING([--enable-gui],
                  [enable_gui=false]
                  )
 
+# Check for termcap and readline or bypass checking for the libraries.
+AC_ARG_ENABLE([libless], [AS_HELP_STRING([--enable-libless],
+                            [use a libreadline replacement with less features])],
+                 [enable_libless=true
```
coolwanglu: This is actually wrong. The last two actions are actually `argument presented` and `argument not presented`, but not `argument enabled` and `argument disabled`.

coolwanglu: I just realized that it derives from `--enable-gui`

coolwanglu:
#101

PR #99, opened 2015-04-24
scanmem: add strdupa() and strndupa() compatibility macros

sriemer:
The macros strdupa() and strndupa() are GNU specific and not
available on other platforms like e.g. the BSD derivatives this way.
The build fails there. We use the GPL and the macros in the glibc
string.h are under LGPL.

So introduce adapted copies of them and use them if they aren't
defined already. This is legal as the GPL license is compatible
and stricter than the LGPL.

Reported-by: Jonathan Pelletier

### DIFF
--- a/scanmem.h
+++ b/scanmem.h
@@ -30,7 +30,7 @@
 #endif
 */
 
-/* provide replacements for GNU macros */
+/* from string.h in glibc */
 #ifndef strdupa
 #define strdupa(s)                                                            \
     ({                                                                        \
--- a/scanmem.h
+++ b/scanmem.h
@@ -6,6 +6,7 @@
 
 #include <stdio.h>
 #include <stdint.h>
+#include <string.h>
 #include <sys/types.h>          /*lint !e537 */
 
 #include "scanroutines.h"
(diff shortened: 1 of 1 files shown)


sriemer:
I've tested this first by using this as is by searching and setting the lives value in Chromium B.S.U.. Then I've undefined the macros and put printf() lines to the macros for checking if they get called. The macros work perfectly.

Funmungus:
I have also tested this with the future Android build, and the replacements work in bionic.

sriemer:
We're both sure about this. So I'll merge this on Monday the latest. Seems like Lu is very busy or on vacation at the moment.
I've also checked the build for my smartphone but I have the glibc there.

sriemer:
The next step would be FreeBSD support. I did a short test but configure failed to read /proc/self/maps. There is no such file or self symlink. But it has a file /proc/$pid/map (so without 's'). By default /proc is not even mounted. So a mount check has to be implemented.

Funmungus:
I see 3 options.  1) Settings file for maps directory, among other possibilities.  2) `AC_DEFINE(PROCMAPS_DIR, [maps])` inside two AC_CHECK_FILE's.  3) In online searches I also see `if test -f "$filename"` which may be better for not failing on cross-compile.

Personally I like settings files.  I also think cross-compiling is only common with embedded software, which usually has its own build environment anyways.

coolwanglu:
I'm back :) I was quite busy with my work stuffs.
I would not worry about license stuffs, since the functions look so simple, I guess no one will be charged for their own implementation of quick-sort.
But in case the code were directly copied from another repo, maybe it's better to note the source.

Funmungus:
Welcome back.

sriemer:
Done. Merge?

coolwanglu:
LGTM  :)

scanmem.h
```diff
@@ -29,6 +30,28 @@
 #endif
 */
 
+/* from string.h in glibc */
+#ifndef strdupa
```
coolwanglu: Oh wait, `#ifndef` does not work for real functions,
so what if `strdupa` is a real function? `#ifndef` will be still true here.

sriemer: At the moment no C library uses real functions for this. We could put a check for these into AC_CHECK_FUNCS() but this can't detect macros and I'm not sure if it can detect inline functions. I'd really avoid making this too complex unless we get a real problem somewhere.

PR #100, opened 2015-04-28
Update handlers.c

calculuswhiz:
I noticed that this caused bytearray and string scans to freeze the GUI. The GUI is expecting a space after the type.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -339,8 +339,8 @@ bool handler__list(globals_t * vars, char **argv, unsigned argc)
         show_error("memory allocation failed.\n");
         return false;
     }
-    char *bytearray_suffix = ", [bytearray]";
-    char *string_suffix = ", [string]";
+    char *bytearray_suffix = ", [bytearray ]";
+    char *string_suffix = ", [string ]";
 
     USEPARAMS();
 


coolwanglu:
In this case, I guess it's better to update the GUI.

calculuswhiz:
Maybe so, but I think consistency in the output would help the GUI in the future. Either way, if the GUI can be fixed itself, I'm ok with that too. It's just kind of inconvenient that it freezes, considering most of what I do is bytearray stuff.
I dunno. Do what you feel is best.

coolwanglu:
What's the consistency you mentioned?

calculuswhiz:
Well, take the following scan result:
<pre>[ 0] 7f22b0d91f70,  5 +        f8f70,  misc, 1337, [I64 I32 I16 ]</pre>
There's a space before the last bracket, right?
On the other hand:
<pre>[ 0] 7f22adab48c4,  1 +          8c4,  misc, fe f7, [bytearray]</pre>
There's no space before the last bracket, which is what the python script is expecting (same is true of string type). (Debugging GC gives me a key error for "bytearray]\n")

Is there an easier way to fix it?

coolwanglu:
Ah, I see, thanks for pointing that out. Let me try to fix it.

sriemer:
This is related to issue #73. If fixing this in the GUI, then we would have to check if spaces are included. If so, use split(). If not, use strip().

sriemer:
Hmm, just appending ".strip()" works for Python 2 but doesn't work for Python 3.
  File "/usr/share/gameconqueror/GameConqueror.py", line 939, in data_worker
    new_value = self.read_value(addr, TYPENAMES_S2G[scanmem_type.split(' ', 1)[0].strip()], cur_value)
KeyError: 'string]\\n'

sriemer:
Another split() for the brace like scanmem_type.split(' ', 1)[0].split(']', 1)[0] should be the solution here.

sriemer:
Hmm, the split works but running in another Python 3 issue with string scanning.

coolwanglu:
`]` should not appear in the string. There must be a bug when parsing the output of scanmem.

coolwanglu:
This [line](https://github.com/scanmem/scanmem/blob/master/gui/GameConqueror.py#L913) is supposed to strip the brackets off the types returned by scanmem. Maybe you can put an assert or print there, to inspect.

After that line, the string should be a number of types separated by spaces.

sriemer:
Thanks for pointing that out. The thing is that these are 16-bit characters and removing one would only remove the '\n'. So it would be even wrong with 8-bit chars. -1. -2 and -3 don't work but with `t = t[1:-4]` it works. `]` and `\n` get removed. I'll do a commit for that and fix the other locations in the string/bytearray search.

sriemer:
Thanks for pointing that out. I've put a print there and checked this. With Python 2 everything works as expected during string search. So we don't have a regression here. But it's still strange to me that this strips both `]` and `\n`. I would expect that only `\n` gets stripped.

I guess @calculuswhiz uses Python 3. The thing is that Python 3 uses 16-bit characters here instead of 8-bit. -1. -2 and -3 don't work but with `t = t[1:-4]` it works for it. `]` and `\n` get removed. I'll do a commit for that and fix the other locations in the string/bytearray search as well. A solution which works for both Python 2 and Python 3 is required.

calculuswhiz:
Oh, I see now. That would make sense.

sriemer:
I've found it. Bytes to unicode conversion is missing for Python 3:
````
diff --git a/gui/GameConqueror.py b/gui/GameConqueror.py
index 9bc8272..fd48a3a 100644
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -905,12 +905,15 @@ class GameConqueror():
             for line in lines:
-                line = str(line)
+                line = str(line.decode("utf-8"))
+                print(line)
                 line = line[line.find(']')+1:]
+                print(line)
                 (a, o, rt, v, t) = list(map(str.strip, line.split(',')[:5]))
                 a = '%x'%(int(a,16),)
                 o = '%x'%(int(o[5:],16),)
                 t = t[1:-1]
+                print(t)
                 self.scanresult_liststore.append([a, v, t, True, o, rt])
             self.scanresult_tv.set_model(self.scanresult_liststore)
 
````

I've searched for the string "Chromium" in Chromium B.S.U..
Python 2:
````
[ 0] 7f0812ed53b0, 29 +        413b0,  heap, Chromium, [string]

 7f0812ed53b0, 29 +        413b0,  heap, Chromium, [string]

string
````

Python 3 before:
````
b'[ 0] 7f0812ed53b0, 29 +        413b0,  heap, Chromium, [string]\n'
 7f0812ed53b0, 29 +        413b0,  heap, Chromium, [string]\n'
string]\n
````

Python 3 after:
````
[ 0] 7f0812ed53b0, 29 +        413b0,  heap, Chromium, [string]

 7f0812ed53b0, 29 +        413b0,  heap, Chromium, [string]

string
````

sriemer:
I've prepared the related patch and continue with the other Python 3 problems. See sriemer@7871e6cd391b64c82a4cf6b6e8cea590e92688d7.

coolwanglu:
@sriemer How about drop support for py2? It could save us from lots of efforts.

sriemer:
@coolwanglu We need to do the changes for Python 3 anyways. It's not hard to support both. Python 2 is what still works best at the moment and as long as there are supported distributions which still come with Python 2 by default I wouldn't want to touch this. I run Ubuntu 14.04 LTS and recently I ran into a problem uninstalling a Python package with Python 3. I had to switch back to Python 2 and then it worked.

coolwanglu:
I see. Hopefully we won't need to pay too much attention on this in the future.

sriemer:
My patches are ready to fix this. So closing this PR.

sriemer:
@calculuswhiz please have a look at PR #102.

PR #101, opened 2015-04-30
fix enable/with arguments in configure.ac

coolwanglu:
fix enable/with arguments in configure.ac; renames libreplacements* to readline*

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,9 +33,9 @@ libscanmem_la_SOURCES = commands.h \
     value.h \
     value.c 
 
-if ENABLE_LIBLESS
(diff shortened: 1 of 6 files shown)


sriemer:
Thanks! Assigning to me for review. I'll look into this ASAP.

Funmungus:
I also just noticed the strange function definitions I put in readline.c.  Sorry I did not look closely enough at that.

sriemer:
LGTM. Thank you very much for cleaning this up!

sriemer:
I'm rebasing my branches to latest upstream. All others should do so as well now. I'm e.g. working on FreeBSD support. I know an IT security professional who would really love having scanmem there.

PR #102, opened 2015-05-03
gui: add support for string/bytearray search with Python 3

sriemer:
Python 3 intentionally doesn't convert from bytes to unicode. This is why the search for string/bytearray freezes in GameConqueror and causes a debug message. So explicitly convert strings returned by scanmem as bytes to unicode.

This has been tested with Python 2.7 and 3.4 and is suitable for 0.15 stable.

Reported-by: calculuswhiz

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -676,20 +676,20 @@ class GameConqueror():
         return None
 
     # parse bytes dumped by scanmem into number, string, etc.
-    def bytes2value(self, typename, bytes):
-        if bytes is None:
+    def bytes2value(self, typename, databytes):
+        if databytes is None:
             return None
         if typename in TYPENAMES_G2STRUCT:
-            return struct.unpack(TYPENAMES_G2STRUCT[typename], bytes)[0]
+            return struct.unpack(TYPENAMES_G2STRUCT[typename], databytes)[0]
         elif typename == 'string':
-            bytes = str(bytes.decode())
-            return repr('%s'%(bytes,))[1:-1]
+            databytes = str(databytes.decode())
+            return repr('%s'%(databytes,))[1:-1]
         elif typename == 'bytearray':
-            bytes = bytes.decode()
-            return ' '.join(['%02x'%ord(i) for i in bytes])
+            databytes = databytes.decode()
+            return ' '.join(['%02x'%ord(i) for i in databytes])
         else:
-            return bytes
-    
+            return databytes
+
     def scan_for_addr(self, addr):
         bits = self.get_pointer_width()
         if bits is None:
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -907,7 +907,7 @@ class GameConqueror():
             # temporarily disable model for scanresult_liststore for the sake of performance
             self.scanresult_liststore.clear()
             for line in lines:
-                line = str(line)
+                line = str(line.decode())
                 line = line[line.find(']')+1:]
                 (a, o, rt, v, t) = list(map(str.strip, line.split(',')[:5]))
                 a = '%x'%(int(a,16),)
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -682,8 +682,10 @@ class GameConqueror():
         if typename in TYPENAMES_G2STRUCT:
             return struct.unpack(TYPENAMES_G2STRUCT[typename], bytes)[0]
         elif typename == 'string':
+            bytes = str(bytes.decode())
             return repr('%s'%(bytes,))[1:-1]
         elif typename == 'bytearray':
+            bytes = bytes.decode()
             return ' '.join(['%02x'%ord(i) for i in bytes])
         else:
             return bytes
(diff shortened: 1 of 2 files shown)


sriemer:
I've tested this with Chromium B.S.U. and searched for "C h r o m i u m" as well as "43 20 68 20" which finds the same value which gets visible in the main menu.

gui/GameConqueror.py
```diff
@@ -905,7 +907,7 @@ def update_scan_result(self):
             # temporarily disable model for scanresult_liststore for the sake of performance
             self.scanresult_liststore.clear()
             for line in lines:
-                line = str(line)
+                line = str(line.decode("utf-8"))
```
coolwanglu: `decode` already returns a `str`, so the outside `str` is not longer useful.

coolwanglu: and `utf-8` is the default, so it could be simplified as `line = line.decode()`

gui/GameConqueror.py
```diff
@@ -681,8 +681,10 @@ def bytes2value(self, typename, bytes):
         if typename in TYPENAMES_G2STRUCT:
             return struct.unpack(TYPENAMES_G2STRUCT[typename], bytes)[0]
         elif typename == 'string':
+            bytes = str(bytes.decode("utf-8"))
             return repr('%s'%(bytes,))[1:-1]
         elif typename == 'bytearray':
+            bytes = bytes.decode("utf-8")
```
coolwanglu: `bytes` is now a built-in type in python3, maybe we should rename it?

gui/GameConqueror.py
sriemer: I agree with `.decode()` without "utf-8" but your other remark is definitely not the case. This is with Python 2 without the `str()`:
````
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/share/gameconqueror/GameConqueror.py", line 889, in scan_thread_func
    self.update_scan_result()
  File "/usr/share/gameconqueror/GameConqueror.py", line 912, in update_scan_result
    (a, o, rt, v, t) = list(map(str.strip, line.split(',')[:5]))
TypeError: descriptor 'strip' requires a 'str' object but received a 'unicode'
````

File: gui/GameConqueror.py (hunk 2 above)
Comment (sriemer) [inline gui/GameConqueror.py]: Would make sense to avoid confusion.

File: gui/GameConqueror.py (hunk 1 above)
Comment (coolwanglu) [inline gui/GameConqueror.py]: Oh, maybe the input is from gtk. I'm not sure. 
So please just leave `str()`

Comment (sriemer) [inline gui/GameConqueror.py]: Maybe the auto-conversion to str() is only valid for Python 3. At the other location I had to add str() for Python 2.

File: gui/GameConqueror.py [hunk 3]
```diff
@@ -681,8 +681,10 @@ def bytes2value(self, typename, bytes):
         if typename in TYPENAMES_G2STRUCT:
             return struct.unpack(TYPENAMES_G2STRUCT[typename], bytes)[0]
         elif typename == 'string':
+            bytes = str(bytes.decode())
```
sriemer: @coolwanglu Would the name `databytes` be okay for you?
````
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -675,19 +675,19 @@ class GameConqueror():
         return None
 
     # parse bytes dumped by scanmem into number, string, etc.
-    def bytes2value(self, typename, bytes):
-        if bytes is None:
+    def bytes2value(self, typename, databytes):
+        if databytes is None:
             return None
         if typename in TYPENAMES_G2STRUCT:
-            return struct.unpack(TYPENAMES_G2STRUCT[typename], bytes)[0]
+            return struct.unpack(TYPENAMES_G2STRUCT[typename], databytes)[0]
         elif typename == 'string':
-            bytes = str(bytes.decode())
-            return repr('%s'%(bytes,))[1:-1]
+            databytes = str(databytes.decode())
+            return repr('%s'%(databytes,))[1:-1]
         elif typename == 'bytearray':
-            bytes = bytes.decode()
-            return ' '.join(['%02x'%ord(i) for i in bytes])
+            databytes = databytes.decode()
+            return ' '.join(['%02x'%ord(i) for i in databytes])
         else:
-            return bytes
+            return databytes
     
     def scan_for_addr(self, addr):
         bits = self.get_pointer_width()
````

coolwanglu: Looks good. Please use it.

sriemer:
Awesome, forcefully pushing works without closing the PR. I've moved the corrections of the decode()s to the respective commits instead of fixing them one or two commits afterwards. I've also rebased the branch to upstream/master so that we are up-to-date again and can continue this one.

sriemer:
@coolwanglu Okay, now I also have the rename. We should be ready to merge now.

sriemer:
I've tested this with both Python versions again, both string and bytearray, both initial and second search. It works.

PR #104, opened 2015-05-04
remove debian/

coolwanglu:
The `debian/` folder was used for my PPA, which is no longer maintained. 
So remove it.
cc @sriemer


sriemer:
Fine for me. But I'll keep my own Debian packaging in my own branches.

coolwanglu:
Oh if you are going to maintain PPA or Debian packages, feel free to add it.

sriemer:
There are less conflicts with upstream if uptream has no debian directory. I like that now. This is likely also better for the Debian package maintainers.

PR #106, opened 2015-05-16
Reduce scanning overhead

mark7:
Eliminate an expensive computation to roughly halve the initial scan
time.

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -271,6 +271,8 @@ bool checkmatches(globals_t * vars,
         total_scan_bytes += tmp_swath_index->number_of_bytes;
         tmp_swath_index = (matches_and_old_values_swath *)(&tmp_swath_index->data[tmp_swath_index->number_of_bytes]);
     }
+    int num_dots = 0;
+    size_t next_dot = total_scan_bytes / 10;
 
     int reading_iterator = 0;
     matches_and_old_values_swath *writing_swath_index = (matches_and_old_values_swath *)vars->matches->swaths;
@@ -342,7 +344,9 @@ bool checkmatches(globals_t * vars,
             --required_extra_bytes_to_record;
         }
 
-        if (EXPECT((total_scan_bytes >= 110) && (bytes_scanned % ((total_scan_bytes) / 10) == 10), false)) {
+        if (bytes_scanned >= next_dot) {
+            num_dots++;
+            next_dot = ((double)total_scan_bytes / 10) * num_dots;
             /* for user, just print a dot */
             show_scan_progress(bytes_scanned, total_scan_bytes);
         }
@@ -484,6 +488,8 @@ bool searchregions(globals_t * vars, scan_match_type_t match_type, const userval
 
         /* load the next region */
         r = n->data;
+        int num_dots = 0;
+        size_t next_r_dot = r->size / 10;
 
 #if HAVE_PROCMEM        
         /* over allocate by enough bytes set to zero that the last bytes can be read as 64-bit ints */
@@ -581,7 +587,9 @@ bool searchregions(globals_t * vars, scan_match_type_t match_type, const userval
             }
 
             /* print a simple progress meter. */
-            if (EXPECT((r->size >= 110) && (offset % ((r->size) / 10) == 10), false)) {
+            if (offset >= next_r_dot) {
+                num_dots++;
+                next_r_dot = ((double)r->size / 10) * num_dots;
                 /* for user, just print a dot */
                 show_scan_progress(bytes_scanned+offset, total_scan_bytes);
             }
--- a/ptrace.c
+++ b/ptrace.c
@@ -266,13 +266,14 @@ bool checkmatches(globals_t * vars,
     long bytes_scanned = 0;
     long total_scan_bytes = 0;
     matches_and_old_values_swath *tmp_swath_index = reading_swath_index;
+    int num_dots = 0;
+    size_t bytes_at_next_dot;
     while(tmp_swath_index->number_of_bytes)
     {
         total_scan_bytes += tmp_swath_index->number_of_bytes;
         tmp_swath_index = (matches_and_old_values_swath *)(&tmp_swath_index->data[tmp_swath_index->number_of_bytes]);
     }
-    int num_dots = 0;
-    size_t next_dot = total_scan_bytes / 10;
+    bytes_at_next_dot = total_scan_bytes / 10;
 
     int reading_iterator = 0;
     matches_and_old_values_swath *writing_swath_index = (matches_and_old_values_swath *)vars->matches->swaths;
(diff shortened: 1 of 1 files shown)


sriemer:
This requires intensive testing to verify that unaligned memory is still found. Memory reads should be much slower than computation. So how did you test this and were you able to remove superfluous memory reads? Or did you just remove important ones? Precise descriptions would help as you're the one who has to make us confident that this won't introduce regressions.

coolwanglu:
I agree with @sriemer .
I wonder if you have measured the running time with and without your patch, not sure how much it can be improved.

mark7:
> This requires intensive testing to verify that unaligned memory is still found.

Why?  The patch should not affect anything other than display of the status meter.

> Memory reads should be much slower than computation.

There may be a platform where this is the case, but at least on my system, a substantial amount of time is definitely being consumed in the computation fo the status update.  This was timed on an i5-4670K @ 3.40GHz.

> So how did you test this and were you able to remove superfluous memory reads?

The only change here is to the computation of the dots in the status.  There are no changes to memory reads (assuming you're referring to in the remote process).

> I wonder if you have measured the running time with and without your patch, not sure how much it can be improved.

Yes, it was run under time(1) on a large program (I don't recall what it was at the moment...I believe Firefox) that with the original code took about thirty seconds, and post-patch requires about eighteen seconds to perform an initial scan.  Both times were fairly stable (though I only did about three runs with each).

coolwanglu:
@mark7 I see, so that's quite good. I didn't suppose this much overhead.
A few comments:
`next_dot` or `next_r_dot` could be confusing. How about `bytes_at_next_not` or other names?
`EXPECT` is supposed to be able to help the compiler predict the branches, please do not remove them.

mark7:
> Defining variables in the middle of a block is bad style.

Apologies.  I noticed reading_iterator next to this this doing the same thing, so I assumed that mid-block definition was fine in the project; will move.

mark7:
> next_dot or next_r_dot could be confusing. How about bytes_at_next_not or other names?

Sure thing.

> EXPECT is supposed to be able to help the compiler predict the branches, please do not remove them.

I'll re-add it for consistency, but I'd be surprised if it produces a measurable difference (at least on x86; I suppose that there could theoretically be an architecture that does very bad branch prediction).

(a) at least as of the Pentium II's branch prediction algorithm (and I doubt that this has changed, since compilers have relied upon this behavior for some time and compiled code targeting it), a conditional forward jump without any cached information about the branch is assumed to be false anyway (if loops will compile to this), and a conditional backwards jump without any cached behavior is assumed to be true (while loops will compile to this).

(b) Your processor will probably cache the result of a branch; I believe that the PII had two bits per and that prediction behavior will be identical to the behavior with EXPECT() macro.  Even if it did no caching and simply exhibited the behavior in (a), you'd get the same prediction behavior, and if it had one bit and  just assumed whatever the last branch was, you'd see ten branch mis-predicts per memory region; not something that'll show up in a measurable way.

coolwanglu:
@mark7 In fact I'm not an expert about `EXPECT`. I just googled as you mentioned this, indeed it seems that `EXPECT` might not be as useful as I had thought of. So it's up to you.
Thanks for the explanation.

mark7:
> Thanks for the explanation.

No sweat, though it's a good fifteen years or so out of date; it's just that branch prediction isn't something that I'd expect to get substantially worse over time.  I'll re-add it, because there are EXPECT() macros throughout the code.

I also obtained real times for you as per request; the reduction was about a 30% improvement, not a 40% improvement as I recalled from above.  Running "echo 10 |time scanmem -p `pidof iceweasel`":

Without patch:

./scanmem -p `pidof iceweasel`  77.12s user 0.60s system 99% cpu 1:17.85 total
./scanmem -p `pidof iceweasel`  77.04s user 0.60s system 99% cpu 1:17.75 total

With patch:

./scanmem -p `pidof iceweasel`  53.44s user 0.58s system 99% cpu 54.108 total
./scanmem -p `pidof iceweasel`  53.14s user 0.60s system 99% cpu 53.828 total

I did some work with prof(1) to identify the earlier hot spot; IIRC there aren't any other really obvious, low-hanging fruit obtainable from one-line changes.

I do have one question, since you're responding so quickly -- I am only a *very* light user of github and normally work with other review systems; I am not familiar with the standard review workflow here.  Is the correct approach (a) to push up a *second* commit onto the wb-speedscan topic branch, and then let the two be squashed when you folks merge, resulting in one commit ultimately pulled into upstream, (b) to generate a new pull request ("wb-speedscan-2") with a new branch name, resulting in one commit to be pulled, (c) to commit --amend on the topic branch and then force-push, and let github infer that it's an update based on the branch name, resulting in one commit pulled into upstream, or (d) to commit a second commit into the topic branch and let both be merged in independently into upstream?

mark7:
I've found more references to (d) than to any other process online, so I'll assume that that is the desired approach: just adding the changes from review as an additional commit on the topic branch and letting both be pulled in.

coolwanglu:
Thanks for the result. Yes it improves the performance a lot.

Regarding the commits, I remember that a final squash is preferred.

The patch now looks good to me :)

@sriemer Any comment?

sriemer:
@coolwanglu I'm still on vacation with no programming environment. This shows that a good commit description is always required. The short as well as the long description are misleading. If people look at `git show`, `git log` and friends they want to catch on the first sight what this is all about. I prefer that all still unreleased fixes get merged like the Linux kernel does it to keep the git repo clean and `git blame` pointing to the right commit. That means closing this PR and opening another one. A long description always has two parts: 1. What's the current situation and what's wrong with it? 2. How to improve the situation? Write this one in imperative mood. People usually link these two with words like 'so' or 'Hence'.

In the code it's not really clear to me why the cast to double is required. Embedded devices aren't that good at floating point operations.

On Monday I'm back at home.

mark7:
> The short as well as the long description are misleading.

All right, I'll rewrite it and explicitly-indicate that it was the status update portion of the scan that was improved, rather than the memory-reading portion.

> I prefer that all still unreleased fixes get merged like the Linux kernel does it to keep the git repo clean and git blame pointing to the right commit.

From a github standpoint, does this mean (a) a rebase that squashes them together and then a repush, or (b) closing the old and opening a new pull request for each patch modification during review?

> In the code it's not really clear to me why the cast to double is required. Embedded devices aren't that good at floating point operations.

If the cast is not performed, you'll have a small amount of inaccuracy: for example, if total_scan_bytes were 123456789, you'd get bytes_at_next_dot for dot 9 being 111111102 instead of 111111110.  This causes output where sometimes the last dot is not printed: for example:

    1344/1437 searching 0x7f0da74fe000 - 0x7f0da7900000............ok
    1345/1437 searching 0x7f0da7900000 - 0x7f0da7c00000............ok
    1346/1437 searching 0x7f0da7d00000 - 0x7f0da8200000...........ok
    1347/1437 searching 0x7f0da8200000 - 0x7f0da8700000...........ok
    1348/1437 searching 0x7f0da9100000 - 0x7f0daa000000...........ok
    1349/1437 searching 0x7f0db0900000 - 0x7f0db0a00000............ok

You'll note that ranges 1346-1348 are missing a dot.

mark7:
Also, re: computational cost, it's worth noting that four lines below the (double) cast is another double cast that does the same computation (long-to-double, floating-point divide, floating-point multiply) that I am doing here.  The double cast this patch performs runs only once per displayed dot; the existing code runs each pass through the loop.

mark7:
As per request, revised patch moved to pull request 107.

PR #107, opened 2015-05-24
Reduce scanning overhead in status update

mark7:
Presently, the status update code frequently performs an expensive
computation.

Restructure the code to reduce the cost of this computation.  This
produces approximately a 30% performance improvement and should have no
functional impact.

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -266,11 +266,14 @@ bool checkmatches(globals_t * vars,
     long bytes_scanned = 0;
     long total_scan_bytes = 0;
     matches_and_old_values_swath *tmp_swath_index = reading_swath_index;
+    int num_dots = 0;
+    size_t bytes_at_next_dot;
     while(tmp_swath_index->number_of_bytes)
     {
         total_scan_bytes += tmp_swath_index->number_of_bytes;
         tmp_swath_index = (matches_and_old_values_swath *)(&tmp_swath_index->data[tmp_swath_index->number_of_bytes]);
     }
+    bytes_at_next_dot = total_scan_bytes / 10;
 
     int reading_iterator = 0;
     matches_and_old_values_swath *writing_swath_index = (matches_and_old_values_swath *)vars->matches->swaths;
@@ -342,7 +345,9 @@ bool checkmatches(globals_t * vars,
             --required_extra_bytes_to_record;
         }
 
-        if (EXPECT((total_scan_bytes >= 110) && (bytes_scanned % ((total_scan_bytes) / 10) == 10), false)) {
+        if (EXPECT(bytes_scanned >= bytes_at_next_dot, false)) {
+            num_dots++;
+            bytes_at_next_dot = ((double)total_scan_bytes / 10) * num_dots;
             /* for user, just print a dot */
             show_scan_progress(bytes_scanned, total_scan_bytes);
         }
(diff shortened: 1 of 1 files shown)


coolwanglu:
Regarding the floating point arithmetic operations. We can fix it in the end, by checking if bytes_at_next_dot is greater than total bytes.

sriemer:
@mark7 Thanks, this looks much cleaner and more understandable now. Also thanks for the explanation regarding the cast to double. This is okay for me this way.

@coolwanglu We have a lot of new contributors now. Isn't it the time to write a documentation for developers for not explaining so many things later in the PRs over and over again?

sriemer:
I also did some profiling on my own with Valgrind and the simple FOSS game Chromium B.S.U. in main menu. I've searched for the i32 lives of the HeroAircraft object in always 31 regions:

````
$ sudo valgrind --tool=callgrind --collect-jumps=yes --dump-instr=yes scanmem -p `pidof chromium-bsu`
0> 4
...
185443> quit
$ kcachegrind callgrind.out.*
````

Yes, the following line is at the top with 5.88 % without the patch:
````
if (EXPECT((r->size >= 110) && (offset % ((r->size) / 10) == 10), false)) {
````
But also this line is at 5.88 %:
````
vars->scan_progress = ((double)bytes_scanned) / total_scan_bytes;
````
I really wondered where the 30 % boost is. Didn't feel that much. With `time` you measure a lot of crap (e.g. how fast you type). A web browser is also too complex for getting reproducible results. In fact your patch improves the situation only by around 5.1 % but that's still quite good as this cost is almost completely removed. You can see that in KCachegrind as it shows a "Total Cycle Estimation Cost" in the status line at the bottom and as Valgrind only accounts the time when the scanmem process is really active. I did each measurement 3 times - always the same result (1 - 7,053,370,288 / 7,433,124,204 = 5.1 %). The scan routines have been called in both cases around 36.4 million times. So numbers are comparable. :-) This is how profiling is done. The only disadvantage is that a tool run with Valgrind can be upto 20 times slower due to all the probe injection.

kcachegrind_scanmem_before
kcachegrind_scanmem_patched

**Comment by coolwanglu** [comment]
@sriemer : @mark7  was using `echo 10|scanmem ...`. Were you able to reproduce it?

**Comment by sriemer** [comment]
@coolwanglu Okay, yes - measured like this with `time` I also get around 33% performance improvement. Profiling seems to be quite intrusive.

**Comment by coolwanglu** [comment]
@sriemer Usually I'm find with variables in the middle of blocks, as long as there are properly scoped. But that's for C++, and I'm not sure about the differences in C actually.

Regarding the documentations, we could try [this](https://github.com/blog/1184-contributing-guidelines), but not sure if people will actually read it.

**Comment by sriemer** [comment]
@coolwanglu The C++ and C99 standards allow variable declarations in the middle of a block. But as long as there is no explicit standard set during compilation I wouldn't rely on that. More freedom often means more chaos. People usually tend to look at the beginning of a block for all variable declarations.
Reference:
http://stackoverflow.com/questions/7218066/does-anyone-know-the-reason-why-the-variables-have-to-be-defined-at-the-top-of-f
Creating a CONTRIBUTING.md file sounds like a good plan to me.

**Comment by sriemer** [comment]
The front-end uses vars->scan_progress. But that's also updated too frequently (several million times). 100 or 1000 samples should be enough. We should have a look at this later on as well.

**Comment by coolwanglu** [comment]
@sriemer Regarding the C standard, probably we can add `-std=c99` or some instructions in README.

And for scan progress, ideally it could be optimized like this:

```
B = bytes_per_dot = ...calculate...
while not finished  
  if bytes left in the current region >= B
    process next B bytes
    print a dot
    B = bytes_per_dot
  else
    B -= bytes left in current region
    finish this region and check next
```

sriemer:
@coolwanglu I've sent you the patch with the rounding handling via email. My method is simpler.

sriemer:
Closing this PR as it should not be merged like it is at the moment. Moved discussion to emails until a final PR can be created.

coolwanglu:
@sriemer Well we may give more time buffer. Meanwhile you can finish your patch and deprecate this one.

sriemer:
This PR is superseded by PR #108.

PR #108, opened 2015-06-01
Optimize scan progress updates

sriemer:
Currently, the scan progress update for the user and the front-end are using expensive computations at each and every scanned byte. This is very inefficient as only 10 dots have to be printed and a progress bar has to be updated while processing a memory region or all already found matches.

This patch set reduces the costs of these computations by replacing them with simpler ones and updating the progress less frequently. Both progress updates are combined to improve the performance even more. Already found matches are split into 100 samples at which the progress for the front-end is reported. At every 10th sample a dot is displayed for the user. During initial scan there is a dot displayed for 10% of each memory region. The progress for the front-end is updated in that moment in time as well and the progress per dot is calculated only per region. All progress updates only use simple addition operations. Also printing a dot is optimized. An inline function is used to get rid of handling of unused parameters and calling a separate function just to print a dot.

The overall performance improvement during initial scan with only few matches is at around **40%** with Intel Core i5 CPUs and at around **50%** with a bit older AMD Phenom II CPUs. Looks like the modern Intel CPUs are better in (floating point) division operations.

The functional change is little and is limited to the progress bar of the front-end. Performance measurement has been done with a test tool just allocating 500 MiB with calloc() and then sleeping. The command to get the real time with only few matches is `time echo 10 | scanmem -p $pid`. It has to be run as root. Profiling has been done with Valgrind and assembly level optimization has been done with `objdump -S .libs/libscanmem.so | less`.

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -68,6 +68,13 @@
 
 /* Progress handling */
 #define NUM_DOTS (10)
+#define NUM_SAMPLES (100)
+#define MAX_PROGRESS (1.0)  /* 100% */
+#if (!NUM_DOTS || !NUM_SAMPLES || NUM_SAMPLES % NUM_DOTS != 0)
+#error Invalid NUM_DOTS to NUM_SAMPLES proportion!
+#endif
+#define SAMPLES_PER_DOT (NUM_SAMPLES / NUM_DOTS)
+#define PROGRESS_PER_SAMPLE (MAX_PROGRESS / NUM_SAMPLES)
 
 /* ptrace peek buffer, used by peekdata() */
 /* make it larger in order to reduce shift */
@@ -257,6 +264,12 @@ bool peekdata(pid_t pid, void *addr, value_t * result)
     return true;
 }
 
+static inline void print_a_dot(void)
+{
+    fprintf(stderr, ".");
+    fflush(stderr);
+}
+
 /* This is the function that handles when you enter a value (or >, <, =) for the second or later time (i.e. when there's already a list of matches); it reduces the list to those that still match. It returns false on failure to attach, detach, or reallocate memory, otherwise true.
 "value" is what to compare to. It is meaningless when the match type is not MATCHEXACT. */
 bool checkmatches(globals_t * vars, 
@@ -269,19 +282,20 @@ bool checkmatches(globals_t * vars,
     long bytes_scanned = 0;
     long total_scan_bytes = 0;
     matches_and_old_values_swath *tmp_swath_index = reading_swath_index;
-    int num_dots = 0;
-    size_t bytes_at_next_dot;
-    size_t bytes_per_dot;
+    int samples_remaining = NUM_SAMPLES;
+    int samples_to_dot = SAMPLES_PER_DOT;
+    size_t bytes_at_next_sample;
+    size_t bytes_per_sample;
 
     while(tmp_swath_index->number_of_bytes)
     {
         total_scan_bytes += tmp_swath_index->number_of_bytes;
         tmp_swath_index = (matches_and_old_values_swath *)(&tmp_swath_index->data[tmp_swath_index->number_of_bytes]);
     }
-    bytes_per_dot = total_scan_bytes / NUM_DOTS;
-    bytes_at_next_dot = bytes_per_dot;
+    bytes_per_sample = total_scan_bytes / NUM_SAMPLES;
+    bytes_at_next_sample = bytes_per_sample;
     /* for user, just print the first dot */
-    show_scan_progress(0, total_scan_bytes);
+    print_a_dot();
 
     int reading_iterator = 0;
     matches_and_old_values_swath *writing_swath_index = (matches_and_old_values_swath *)vars->matches->swaths;
@@ -294,6 +308,7 @@ bool checkmatches(globals_t * vars,
 
     int required_extra_bytes_to_record = 0;
     vars->num_matches = 0;
+    vars->scan_progress = 0.0;
     
     if (choose_scanroutine(vars->options.scan_data_type, match_type) == false)
     {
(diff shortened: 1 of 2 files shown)


sriemer:
Tool to provide a huge amount of memory:
````
#include <stdlib.h>
#include <unistd.h>

int main(void)
{
    long *data = (long *)calloc(1, 500 * 1024 * 1024);

    if (!data)
        return -1;
    while (1)
        sleep(1);
    return 0;
}
````

**Comment by sriemer** [comment]
The correctness of all loop iterations has been tested by printing the number of bytes scanned or the offset within a region at each sample/dot. Also the scan progress for the front-end has been printed after each update to check if it's correct.

**Comment by coolwanglu** [comment]
Hmm, I just browsed through GUI files, it seems that the dots are no longer used, as the scan progress is obtained via a function.
And regarding the variable for progress, if `double` is too slow, we could just integer for percentages so far.

**Comment by sriemer** [comment]
@coolwanglu I've tried the integer approach with 10000 being `MAX_PROGRESS` and 100.00% but as it requires additional multiplication (`bytes_per_dot * MAX_PROGRESS / total_scan_bytes`) per region and conversion to double for the GUI again, there hasn't been a performance gain. It was exactly the same. Only when using the GUI it was even slower. I've checked it with a stop watch: Before it took 30s to search for '10' in 500 MiB zeroed memory and with the `scan_progress` as integer it took 35s due to the conversions to double for the GUI in `get_scan_progress()`.
Btw.: Using the GUI also takes **triple the time** (30s) than using scanmem directly (10.7s on Intel Core i5).

**Comment by coolwanglu** [comment]
I meant to change `get_scan_progress` to return integers between 0 and 100. We could even use binary values like 0~127 to speed up if necessary.

**Comment by sriemer** [comment]
Sounds like a good idea but will take some extra time and testing. As the GUI is involved in this, I could imagine doing that in a third patch. The focus of the other two patches is speeding up scanmem.
What do you say about these before thinking about improving more stuff apart from this scope?

**Comment by sriemer** [comment]
Benchmark results:
<pre>
AMD Phenom II

                     Time in s                      Speedup
Pattern    unpatched  patch 1  patch 1+2    patch 1  patch 2  patch 1+2
10           26.7       19.4     13.6       27.34%   29.90%    49.06%
0\n10       131.1      117.5    108.2       10.37%    7.91%    17.47%
0            49.4       44.1     38.5       10.73%   12.70%    22.06%
10 after 0   81.7       73.4     69.7       10.16%    5.04%    14.69%

Intel Core i5

                     Time in s                      Speedup
Pattern    unpatched  patch 1  patch 1+2    patch 1  patch 2  patch 1+2
10           18.0       11.7     10.7       35.00%    8.55%    40.56%
0\n10       104.9       94.4     93.5       10.01%    0.95%    10.87%
0            35.0       31.5     31.5       10.00%    0.00%    10.00%
10 after 0   69.9       62.9     62.0       10.01%    1.43%    11.30%
</pre>

Intel definitely uses some (floating point) division magic as the effect of the second patch isn't that big there. But the AMD system shows that there is a quite noticeable effect on less sophisticated systems. For embedded systems the effect is very likely even bigger.

**Comment by sriemer** [comment]
I've just checked having no progress handling at all on the Intel Core i5 and I get similar values than with both patches. So there is no further improvement possible regarding the progress handling. This is also what Valgrind profiling shows.

**Comment by sriemer** [comment]
@coolwanglu The GUI is just querying the progress too frequently. What about doing that just every 100ms?
All it takes is a tiny `time.sleep(.1)`:
````
@@ -750,6 +750,7 @@ class GameConqueror():
 
     # this callback will be called from other thread
     def progress_watcher(self):
+        time.sleep(.1)
         Gdk.threads_enter()
         self.scanprogress_progressbar.set_fraction(self.backend.get_scan_progress())
         Gdk.flush()
````

File: ptrace.c [hunk 1]
```diff
@@ -372,9 +405,9 @@ bool checkmatches(globals_t * vars,
     /* TODO: we'll need progress for checkmatches too */
     if (vars->options.backend == 1)
     {
-        show_scan_progress(total_scan_bytes, total_scan_bytes);
+        print_a_dot();
```
coolwanglu: I think this if block can be removed.

ptrace.c
```diff
@@ -598,9 +642,9 @@ bool searchregions(globals_t * vars, scan_match_type_t match_type, const userval
     }
 
     /* tell front-end we've done */
-    if(vars->options.backend == 1)
-        show_scan_progress(total_scan_bytes, total_scan_bytes);
-    vars->scan_progress = 1.0;
+    if (vars->options.backend == 1)
```
coolwanglu: This could be removed also.

coolwanglu:
@sriemer Yes of course, in fact I'm a little surprised that I had not done so.
Could you put `sleep` at the end of the function?

The PR looks good to me, do you want to push another commit for `sleep`?

sriemer:
@coolwanglu Thanks, yes, I'd like to do one more commit to remove the if conditions for the GUI which are just used to print unneeded dots and a further commit to add the sleep for 100ms at the end of the `progress_watcher` function.

coolwanglu:
@sriemer Oh wait, I see that the `watcher` routine is called via `gobject.idle_add`, that's why I did not add a timer there. I guess we should use `gobject.timeout_add` with a proper interval, and the callback function should always return TRUE.

See: http://www.pygtk.org/pygtk2reference/gobject-functions.html

sriemer:
@coolwanglu Agreed, also the priority can be lowered to idle again. With the interval in ms, we just have to put 100 there. I already feared that my quick approach to just test if we waste precious compute time there is too hacky to become the final fix.

coolwanglu:
In fact I think `gobject.idle_add` is good, it's just python that does not support multithreading well.

sriemer:
@coolwanglu This PR can be merged now IMHO.

coolwanglu:
Sure. In fact I do not receive notifications for new commits.

Issue #103, opened 2015-05-03
gui: hexview doesn't work properly with Python 3


sriemer:
The hexview doesn't open with Python 3. I'm not sure yet how to fix it. It shows the following debug message:

````
Traceback (most recent call last):
  File "/usr/share/gameconqueror/GameConqueror.py", line 317, in MemoryEditor_Button_clicked_cb
    self.browse_memory()
  File "/usr/share/gameconqueror/GameConqueror.py", line 743, in browse_memory
    self.memoryeditor_hexview.payload=data
  File "/usr/share/gameconqueror/hexview.py", line 646, in set_payload
    view.render(self._payload)
  File "/usr/share/gameconqueror/hexview.py", line 440, in render
    " ".join(map(convert, txt[i * bpl:(i * bpl) + bpl]))
  File "/usr/share/gameconqueror/hexview.py", line 431, in <lambda>
    convert = lambda x: '%02X'%(ord(x),)
TypeError: ord() expected string of length 1, but int found
````

coolwanglu:
I think it's still related to `bytes` and `str` in python3.
Although I don't have a GUI now, by reading the source code, it turns out that the memory data is read from a file, which was `str` in python2, and we format each char using `'%02X'%(ord(c),)` for each `c`.

Now we want to do this in python3, but `c` is now an integer, as the data read from a file is `bytes`.

coolwanglu:
One way of solving this, although not beautiful:
```covert = lambda x: '%02X'%(ord(x),) if sys.version >= '3' else lambda x: '%02X'%(x,)`

sriemer:
Your proposal doesn't work. The decoding of the payload to utf-8 also doesn't work. But the working solution is described here:
http://python3porting.com/noconv.html#more-bytes-strings-and-unicode

Adding the following to misc.py:
````
import sys
import codecs

PY3K = sys.version_info >= (3, 0)

# for Python 3: convert to unicode just like Python 2 does
def u(x):
    if PY3K:
        return codecs.unicode_escape_decode(x)[0]
    else:
        return x
````
Changing GameConqueror.py:
````
from misc import u
self.memoryeditor_hexview.payload = u(data)
````
The data representation is correct this way but it still crashes with Python 3 when clicking at the right side of the hexview where the ASCII chars are displayed.

````
RuntimeError: maximum recursion depth exceeded
Fatal Python error: Cannot recover from stack overflow.
error: 
Killed by signal 6.
````
[...]

Issue #114, opened 2015-06-14
gui: hexview: hex/ascii text gets edited at a too low position with Python 3


sriemer:
Python 3 explicitly doesn't convert division results from float to int. This is why too low positions get edited in the hexview hex or ascii text with Python 3. So integer conversion is required for the calculation of `org_off` in the __on_key_press() methods.

sriemer:
This issue occurs only after fixing #103. The patch is in place in the PR. So closing this for now.

Issue #113, opened 2015-06-12
gui: hexview: cursor movement is incorrect (Python 2 and 3)


sriemer:
Both methods `HexText::__on_move_cursor()` and `AsciiText::__on_move_cursor()` have the same problems: The cursor can't be moved backwards and when moving a line down, then the cursor is also moved one position forward. This makes cursor control unusable.

A rework of these functions is required.

sriemer:
The fix is merged.

PR #111, opened 2015-06-08
value/gui: rework error handling in and around valtostr()

sriemer:
When searching for numbers with the differential search (?, <, >, =),
the GUI freezes at the point when it should list the scan results
with Python 2 and 3. Analysis has shown that it happens if "unknown"
or a number as big as the buffer are printed without a type. The
printing to the buffer is completely unchecked. All this has been
touched by commit 1f34f3e7 ("implemented dataarray representation")
the last time back in 2010.

So check if the printing to the buffer results in valid output and
make sure that there is always a type printed. For the error case use
"unknown, [unknown]" to print a type as well. Move the error handling
completely into valtostr() and change the return type to void as it
is the same for all call locations.

If the type is unknown, then the GUI can't and shouldn't process such
results further. So ignore such lines coming from the scanmem 'list'
command.

This has been tested with Python 2.7 and 3.4, with values which were
too long before, with values which were reported as "unknown" without
a type, with the regular number, bytearray and string search again for
not introducing any regressions.

Reported-by: Geoff Maciolek
Fixes: GitHub issue #109

PR #115, opened 2015-06-15
gui: fix the Hexview (esp. for Python 3)

sriemer:
With Python 2 the cursor movement is incorrect. The cursor can't be moved backwards and when moving a line down, then the cursor is also moved one position forward. This makes cursor control unusable. With Python 3 the hexview window even doesn't open at all and a Python debug message is printed instead as the payload data is in bytes and not in unicode. There is a recursion problem on ascii/hex change and conversions of division results to integer are missing.

All this is fixed by this pull request so that this behaves the same with Python 3 as with Python 2 and the hexview becomes actually usable.

The following issues are fixed: #103, #113 and #114.

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -39,6 +39,7 @@ from gi.repository import GObject
 from consts import *
 from hexview import HexView
 from backend import GameConquerorBackend
+from misc import u
 import misc
 
 import locale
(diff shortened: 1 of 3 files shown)


sriemer:
This is suitable for 0.15 stable.

sriemer:
@GeoffMaciolek Would be cool if you could test this as well to make Lu more confident that this will work.

coolwanglu:
Thanks. LGTM.
Just curious why #103 occurs with py3?

sriemer:
The recursion problem? I guess Python 2 has a detection for that and doesn't complain as functionality was exactly the same when removing this code. I have no idea for what it was intended for. Can't see that in your short commit descriptions. ;-) Putting int conversions there for Py3 didn't help. The hex and ascii code changes upon init and when scrolling a line down/up at the bottom/top. But it crashed when clicking at the ascii text.

sriemer:
Thanks for merging. :-) Do you proceed with PR #111?

Issue #120, opened 2015-07-12
gui: bytearray error - can't convert 0xfe to utf-8


calculuswhiz:
Hi, again.
I'm getting an error in GameConqueror when I scan bytecodes. It says:
```
Traceback (most recent call last):
  File "/usr/share/gameconqueror/GameConqueror.py", line 947, in data_worker
    new_value = self.read_value(addr, TYPENAMES_S2G[scanmem_type.split(' ', 1)[0]], cur_value)
  File "/usr/share/gameconqueror/GameConqueror.py", line 974, in read_value
    return self.bytes2value(typestr, self.read_memory(addr, self.get_type_size(typestr, prev_value)))
  File "/usr/share/gameconqueror/GameConqueror.py", line 689, in bytes2value
    databytes = databytes.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfe in position 0: invalid start byte
```
Strings work fine now, though, which is good.

sriemer:
Sounds familiar. Need to replace all .decode() calls with u() calls like already done for the hexview.

sriemer:
Please test the code from PR #121 as well.

sriemer:
The fix has been merged.

calculuswhiz:
Works great! Thanks.

Issue #116, opened 2015-06-23
no python cache files generated/deployed for gameconqueror via make install target


anthraxx:
Hey, the issue is more problematic (for packaging) then it sounds, first let me explain it:

after 'make install' the python files will be distributed to 'usr/share/gameconqueror/' without generating .pyc and .pyo files. If the GUI is started with uid=0 then python has write access to that directory and will generate those caching files for all modules that are somewhere imported (via GameConqueror.py). This results in dirty and untracked files laying around in that directory. If the software is then uninstalled, those cache files remain as they are not tracked as packaged files.

example for python3:
>
find /usr/share/gameconqueror/ -name "*.pyc" -o -name "*.pyo"

>
/usr/share/gameconqueror/__pycache__/backend.cpython-34.pyc
/usr/share/gameconqueror/__pycache__/consts.cpython-34.pyc
/usr/share/gameconqueror/__pycache__/misc.cpython-34.pyc
/usr/share/gameconqueror/__pycache__/hexview.cpython-34.pyc

Resolution:
generate those in a Makefile target and deploy them via 'install' target:

Proposal 1:
add those to make target and distribute the .pyc and .pyo files
>
python -m compileall .
python -O -m compileall .

Proposal 2::
better way for python via setuptools setup.py
> python setup.py install -O1 

packages that distribute python files/modules that are imported should always by itself create and deploy the caching files to avoid cluttering the filesystem and having untracked files laying around.

sriemer:
Sounds reasonable to me. Thanks for reporting! Should be fixed ASAP but doesn't require a new upstream release. Package maintainers can apply the patch on top of the release.

Unfortunately, Python is my weak spot. I'm not sure yet how to do that with autotools. Would you be able to provide a patch faster?

anthraxx:
I'm not a big fan of autotools but I can look into this in the following days. Which way do you prefer the solution, should I implement a full setup.py with setuptools or do you only want to issue the python compileall modules?

If you don't mind, I would go the setuptools way and provide a patch that adds setup.py and calls it via autotools.

sriemer:
If that's the common and preferred method, then I also prefer the setuptools way. I have to do some searching on the web.

sriemer:
No progress here? Hmmm, then I think I have to do it myself somehow. In my Python book there are 12 pages about distributing Python projects with `distutils`.

anthraxx:
sorry I'm very busy in the past weeks, no time for additional tasks currently :(
[...]

PR #121, opened 2015-07-13
v0.15.2 Python fixes: Python cache + unicode conversion

sriemer:
The Python handling for the GUI is non-standard. This is why the
Python cache isn't generated during the installation. It is
generated when calling gameconqueror instead and causes files to
stay after uninstalling GameConqueror. This is unclean. The
preferred GNU automake way is the following:
http://www.gnu.org/software/automake/manual/html_node/Python.html

Add AM_PATH_PYTHON() to detect the Python interpreter and switch
over to 'python_PYTHON' and 'pythondir' to let automake generate
the Python cache upon installation. But GameConqueror.py needs to
become executable again. Use the 'install-data-hook' and 'chmod'
for that.

Reported-by: @anthraxx
Fixes: GitHub issue #116

Converting strings to utf-8 with the .decode() method has the
problem that not all characters (like e.g. 0xfe) can be converted
to unicode. But with the u() function in misc.py which has been
introduced for the hexview we have a function which converts with
Python 3 to unicode just like Python 2 does. So use it everywhere
to convert to unicode.

This has been tested with Python 2.7 and 3.4.

Reported-by: @calculuswhiz
Fixes: GitHub issue #120

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ config.status
 config.sub
 libtool
 ltmain.sh
+py-compile
 scanmem
 stamp-h1
 tags
--- a/configure.ac
+++ b/configure.ac
@@ -115,6 +115,7 @@ AC_ARG_ENABLE(gui, [AS_HELP_STRING([--enable-gui],
                             [enable gameconqueror, the gui front-end of scanmem])])
 AM_CONDITIONAL([ENABLE_GUI], [test "x$enable_gui" = "xyes"])
 AS_IF([test "x$enable_gui" = "xyes"], [
+  AM_PATH_PYTHON([2.7])
   AC_CONFIG_FILES([
     gui/Makefile
     gui/icons/Makefile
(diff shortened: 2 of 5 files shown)


sriemer:
TODO: Add the `Reported-by` lines to the commit descriptions and forcefully push.
Contributors have to be credited to motivate them to contribute again.
Sorry, I was in a hurry. Rather spent more time on testing these changes. ;-)

sriemer:
TODO: rename
````
treewide: use u() instead of .decode() for unicode conversion
````
to
````
gui: use u() instead of .decode() for unicode conversion
````
as it just affects the GUI.

sriemer:
TODOs done, forcefully pushed new commit descriptions.

coolwanglu:
LGTM. Thanks!

sriemer:
Thanks! This was fast this time. :-)

Issue #122, opened 2015-07-15
maps: ASLR/PIE: 'exe' region type detection doesn't work with latest kernels


sriemer:
What worked with kernel 3.13.0-55 in Ubuntu 14.04 Trusty doesn't work with the 3.13.0-57 kernel any more. The reason is that a lot of memory regions are loaded between .text and .rodata - even an executable one. The current region type detection in `maps.c` can't handle that, the type becomes "misc" instead of "exe" and the wrong load address is determined. This way also the found match offset is wrong and the found value can't be put into a `ugtrain` config for static memory cheating with ASLR/PIE.

The already configured match offsets for static memory cheating with PIE which have been discovered with kernel 3.13.0-55 still work with kernel 3.13.0-57 and `ugtrain`. `ugtrain` just searches for the .text region to determine the load address of the executable. So only `scanmem` is affected.

Example: "Power" value in Warzone 2100 stored in .bss

````
7fa8a979c000-7fa8a9ac0000 r-xp 00000000 08:13 1178291                    /usr/games/warzone2100
7fa8a9ac9000-7fa8a9ad1000 rw-s 00000000 00:04 74219                      /drm mm object (deleted)
7fa8a9ad1000-7fa8a9ad6000 rw-s 00000000 00:04 72406                      /drm mm object (deleted)
7fa8a9ad6000-7fa8a9ae6000 rw-s 1043eb000 00:05 10404                     /dev/dri/card0
7fa8a9ae8000-7fa8a9ae9000 rwxp 00000000 00:00 0 
7fa8a9ae9000-7fa8a9af1000 rw-s 00000000 00:04 74207                      /drm mm object (deleted)
7fa8a9af1000-7fa8a9af9000 rw-s 00000000 00:04 74206                      /drm mm object (deleted)
7fa8a9af9000-7fa8a9afa000 rw-s 00000000 00:04 74203                      /drm mm object (deleted)
7fa8a9afb000-7fa8a9b03000 rw-s 00000000 00:04 74202                      /drm mm object (deleted)
7fa8a9b03000-7fa8a9b29000 rw-p 00000000 00:00 0 
7fa8a9b2b000-7fa8a9b3d000 rw-p 00000000 00:00 0 
7fa8a9b3d000-7fa8a9b5d000 rw-s 104ccf000 00:05 10404                     /dev/dri/card0
[...]

PR #124, opened 2015-07-19
maps: handle the gap behind .text of the exe

sriemer:
With the 04/2015 Linux kernel commit a87938b2e246
("fs/binfmt_elf.c: fix bug in loading of PIE binaries")
we've noticed that our assumption that nothing gets loaded into
the gap behind the .text region of the executable has proven wrong.
Unrelated regions might be loaded via mmap() to that location.
Scanmem sets the region type to 'misc' instead of 'exe' for the
.rodata, .data and .bss sections of the executable and also the
load address is incorrect this way.

So count the regions belonging to the executable separately and if
there are not at least two of them, don't reset that count and try
to detect further regions belonging to the executable. There must
be at least a .data region. Also remember the load address of the
executable and assign it to the other regions belonging to it.

This has been tested with PIE binaries with and without an affected
kernel. This has been also tested with regular executables without
PIE with and without an affected kernel.

Reference:
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=a87938b2e246b81b4fb713edb371a9fa3c5c3c86

Fixes: GitHub issue #122

### DIFF
--- a/maps.c
+++ b/maps.c
@@ -4,6 +4,7 @@
  Copyright (C) 2006,2007,2009 Tavis Ormandy <taviso@sdf.lonestar.org>
  Copyright (C) 2009           Eli Dupree <elidupree@charter.net>
  Copyright (C) 2009,2010      WANG Lu <coolwanglu@gmail.com>
+ Copyright (C) 2014,2015      Sebastian Parschauer <s.parschauer@gmx.de>
 
  This program is free software; you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
@@ -48,9 +49,9 @@ bool readmaps(pid_t target, list_t * regions)
     char name[128], *line = NULL;
     char exelink[128];
     size_t len = 0;
-    unsigned int code_regions = 0;
+    unsigned int code_regions = 0, exe_regions = 0;
+    unsigned long prev_end = 0, load_addr = 0, exe_load = 0;
     bool is_exe = false;
-    unsigned long prev_end = 0, load_addr = 0;
 
 #define MAX_LINKBUF_SIZE 256
     char linkbuf[MAX_LINKBUF_SIZE], *exename = linkbuf;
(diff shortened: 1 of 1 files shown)


sriemer:
@coolwanglu We really have to fix this in scanmem. Due to the gap between .text and .rodata there is always the chance that something else mmap()s unrelated regions to this location.

maps.c
```diff
@@ -119,7 +119,13 @@ bool readmaps(pid_t target, list_t * regions)
                  * There should never be more than these four regions.
                  * The data regions use their variables relative to the load
                  * address. So determining it makes sense as we can get the
-                 * variable address used within the binariy with it.
+                 * variable address used within the binary with it.
+                 *
+                 * 04/2015: Usually there is nothing in between these four
```
coolwanglu: Instead of appending the log, I suggest you to update the existing ones to reflect the current situation.

maps.c
```diff
@@ -131,19 +137,36 @@ bool readmaps(pid_t target, list_t * regions)
...
+                        if (strncmp(filename, exename, MAX_LINKBUF_SIZE) == 0) {
+                            exe_regions = 1;
+                            exe_load = start;
                             is_exe = true;
+                        }
+                        strncpy(binname, filename, MAX_LINKBUF_SIZE);
+                        binname[MAX_LINKBUF_SIZE - 1] = '\0';  /* just to be sure */
+                    } else if (exe_regions == 1 && filename[0] != '\0' &&
+                      strncmp(filename, exename, MAX_LINKBUF_SIZE) == 0) {
+                        /* Kernel with 04/2015 PIE loading bugfix - aka ugly PIE */
```
coolwanglu: Why 'ugly PIE' is aka here, I cannot find any reference.

coolwanglu:
Just finish reading the relevant text. Frankly speaking I don't 100% understand our algorithm there. It seems that some of our assumptions there were broken by the latest Linux kernel.

I think test cases may be introduced to test this function, just to make sure that its behavior won't change if updated in the future.

maps.c
sriemer: Okay, can be removed.

sriemer:
Just the assumption that nothing else can be loaded between .text and .rodata of the executable has been broken. The gap there is intended behavior. The Linux kernel puts `mmap()`ed regions into that gap.
How do you imagine test cases?
There are no other test cases. It depends on the Linux kernel and tools which use `mmap()`.

The algorithm works as before. After reading the .text region an unrelated region is read marking the end of the executable. But as we count the regions belonging to the executable now and know that executables without .data don't exist, we check the filename for all unrelated regions loaded into the gap. If we find a second region belonging to the executable by filename, then we continue with the exe processing as if there wouldn't be anything else in the gap.

coolwanglu:
@sriemer a test case would be a txt file with the content of an actual `maps` file we have ever seen. And we should except the function to return the correct load address and number of regions for example.

Could be easier to maintain this way. But it's up to you, I'm thinking about this because I found it not easy to prove the correctness by reading the source code.

sriemer:
My test case would be usually taking a game which uses static memory with PIE, finding the variable with scanmem and lock it with ugtrain. That's what I do e.g. with Warzone 2100.

coolwanglu:
@sriemer All right if you find it easier to test it that way.

It LGTM, feel free to merge it after updating the comments.

Thanks!

sriemer:
Thanks! Yes, static memory cheating would be the common use case for it. Otherwise, you wouldn't have to think about the region type detection at all. Doing it with a real world example tests multiple things at once and you read from the actual /proc/$pid/maps file and don't have to maintain a special case where you read from a different file.

Would it be okay for you if I add my copyright for last year and this year as well to that file?
The region type detection is quite unique and worth being under copyright.

coolwanglu:
Sure!

sriemer:
@coolwanglu Is it better now? I've only changed the description. I've created a branch "old-pie-fix" in case you want to compare.

coolwanglu:
Looks good now. Thanks!

PR #127, opened 2015-09-16
Fix XDG_RUNTIME_DIR issues for Ubuntu

fizzfaldt:
In Ubuntu 14.04 with mate (possibly others),
not exporting DISPLAY and XAUTHORITY causes gameconqueror to fail to
load with errors like:
   error: XDG_RUNTIME_DIR not set in the environment
    (GameConqueror.py:4181): Gtk-CRITICAL **: gtk_clipboard_get_for_display:
        assertion 'display != NULL' failed

This change allows gameconqueror to work again in Ubuntu 14.04

### DIFF
--- a/gui/gameconqueror.in
+++ b/gui/gameconqueror.in
@@ -3,7 +3,7 @@
 DATADIR=@PKGDATADIR@
 
 if [ -e "/usr/bin/pkexec" ]; then
-    exec /usr/bin/pkexec $DATADIR/GameConqueror.py
+    exec /usr/bin/pkexec env DISPLAY=$DISPLAY XAUTHORITY=$XAUTHORITY $DATADIR/GameConqueror.py
 else
     echo "install policykit!"
 fi


sriemer:
Thanks for bringing this to upstream like requested! Looks good to me!
But we use the prefix "gui: " in the short description to show that this is a patch regarding the GUI. 

The long description has more than 70 chars.
Could you move "gtk_clipboard_get_for_display:" into a separate line?

On Ubuntu 14.04 with the Unity desktop this issue doesn't exist.
Could you change the last sentence into "So always export DISPLAY and XAUTHORITY."?

The second part of the long description should be always written in imperative mood as if you would give orders to the code base to change its behavior.

You can do the following sequence to replace the patch without closing this PR:
````
git format-patch HEAD^1
<your_editor> 0001-*     # edit the commit description
git reset --hard HEAD^1
git am 0001-*                 # apply reworked patch
git diff origin <branch_name>   # check if everything is fine
git push -f origin <branch_name>   # forcefully push to replace the patch but not closing the PR
````

I'll test this patch on Ubuntu 14.04 with the Unity desktop when today in the evening.

fizzfaldt:
I'll try to get to it by this weekend, thanks for the contents

sriemer:
Works. Thanks!
I've reworked the patch myself and applied it directly. This is trivial enough for not requiring any merge commits. The exec line became a bit long. So I've wrapped it and I've also changed the short description to something useful.

12345ieee:
So, I've unknowingly worked on reversing this, I found out the commits only later.

Point is, the correct way to do it should be using the policy (I mean, GParted, Synaptic and libvirt do it this way, I refuse to believe they are all wrong) and not forcibly passing the envvars, the man page at https://www.freedesktop.org/software/polkit/docs/0.105/pkexec.1.html does not even mentions the `env` "hack".

Consider that you already ship a policy (which is i18n-able and customizable), that you actually never use right now, as `pkexec` is called on `env`, rather than directly on `gameconqueror`.

Considering that Debian raised a bug specifically for the pkexec stuff ( see #149 ) maybe it's worth to look into it.

I've tested the change in Debian testing both under GNOME and MATE and everything seems to work perfectly.

The changes are in commits d377514deed582a540811d4316ac3fee63330d28  and d2567b42e75f229d9915c3bdb1a3a5cf723cdbf4.

sriemer:
Thanks for bringing this to my attention again. I also noticed an issue on openSUSE Leap 42.1 requiring to revert this. But I think there is no need to change the hard-coded path to pkexec unless a distro would put it somewhere else.

12345ieee:
The `pkexec` path is just a paranoid thought of mine, for example `gparted` just calls `pkexec` without even bothering to check if it's there, but, IMHO, if you check that something exists, do it well or don't do it altogether.

As for the problem @fizzfaldt was having, maybe it was the `exec` that was fixed afterwards.

Do you want a PR/patches or will you fix this yourself?

sriemer:
Thanks, I've reworked your second patch a bit and will do another one to just call `pkexec` letting the shell look up the path.

sriemer:
@12345ieee: I've applied your patches with minor changes. Thanks!
Please don't touch the manpage scanmem.1. I'm completely reworking/updating it right now.

12345ieee:
To be even more paranoid we could have put `#!/bin/sh`in the shebang, but if someone has no `bash` they are really asking for trouble.

PR #139, opened 2015-10-14
Android Makefiles, autoconf, and build script

Funmungus:
This set of changes is a combination of my work and work done by JianXiong Zhou.

Files in the jni/ directory allow building with the ndk-build command.  config.h must first exist by executing `./configure --without-readline`
build_for_android.sh: Automate the Android build process.  To build with libreadline the `NDK_STANDALONE_TOOLCHAIN` variable must be set.  It is suggested to also have the libreadline and ncurses sources set into the `LIBREADLINE_DIR` and `NCURSES_DIR` variables, respectively.
Please execute `./build_for_android.sh help` for more information on Android builds.

getline: This modified version from JianXiong Zhou works well with the readline replacement.  All commands can be parsed without history.
A limitation of ndk-build and Android.mk is that I was not able to link libreadline and ncurses built with a NDK standalone toolchain.
A limitation of the NDK standalone toolchain build is that lines between `if ANDROID_HOST... endif` have to be commented out for versions of Android below 5.0.
Additionally, Darwin support has been added to the autogen and build_for_android scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,13 @@ po/POTFILES
 po/ja.gmo
 po/ru.gmo
 po/stamp-it
+# Android builds
+obj/
+libs/
+# libreadline and ncurses downloads and links
+readline-6.3.tar.gz
+readline-6.3
+ncurses-6.0.tar.gz
+ncurses-6.0
+readline
+ncurses
--- a/README
+++ b/README
@@ -45,6 +45,10 @@ GameConqueror is a GUI front-end for scanmem, providing more features, such as:
 
 See gui/README for more detail.
 
+## Android
+You need a [standalone toolchain of Android NDK](https://developer.android.com/ndk/guides/standalone_toolchain.html#itc) (the Advanced method) to build interactive capabilities for Android.  For more information, run:
+
+  ./build_for_android.sh help
 
 ## License: 
 
--- a/build_for_android.sh
+++ b/build_for_android.sh
@@ -1,0 +1,159 @@
+#!/bin/sh
+
+if [ "x$1" = "x--help" ] || [ "x$1" = "xhelp" ] || [ "x$1" = "x-h" ]; then
+  echo "## Before building:"
+  echo "    # run autogen.sh if \"./configure\" does not exist"
+  echo "    ./autogen.sh"
+  echo
+  echo "## To build without an NDK standalone toolchain, and without libreadline:"
+  echo "    # For given NDK directory <NDKPath>, execute from the scanmem source directory:"
+  echo "    ./configure --without-readline"
+  echo "    <NDKPath>/ndk-build>"
+  echo "or"
+  echo "    export NDK_HOME=\"<NDKPath>\""
+  echo "    ./build_for_android.sh"
+  echo
+  echo "## To build with standalone toolchain:"
+  echo "    export NDK_STANDALONE_TOOLCHAIN=\"/your/toolchain/path\""
+  echo "    export HOST=\"your-androideabi\" # Default arm-linux-androideabi"
+  echo "    ./build_for_android.sh"
+  echo
+  echo "## Advanced features and Environment variables that may be set..."
+  echo "NDK_STANDALONE_TOOLCHAIN - A standalone toolchain is required to build full capabilities."
+  echo "NDK_HOME                 - This will allow to use the ndk-build, and a full standalone toolchain is not required.  If this is set, the scanmem executable will be built with ndk-build, and not the standalone toolchain.  This needs to be used if the target Android version is below 5.0.  This requirement may be fixed in the future."
+  echo "ANDROID_NDK_HOME         - Same as NDK_HOME"
+  echo "HOST                     - Compiler architecture that will be used for cross-compiling, default is arm-linux-androideabi"
+  echo "SCANMEM_HOME             - Path which has scanmem sources, and will be used to build scanmem.  Default current directory"
+  echo "LIBREADLINE_DIR          - Path which has libreadline sources: Default is to download sources"
+  echo "NCURSES_DIR              - Path which has ncurses sources: Default is to download sources"
+  echo
+  exit 0
+fi
+
+# Build with Android.mk
+# Do from scanmem directory, or SCANMEM_HOME
+ANDROID_MK(){
+  # Readline support with ndk-build does not work
+#  if [ -f ./jni/libreadline.a ] && [ -f ./jni/libncurses.a ]; then
+#    export WITH_READLINE=1
+#    export HAVE_LIBREADLINE=1
+#  fi
+  local ndkhome
+  if [ ! "x${NDK_HOME}" = "x" ]; then
+    ndkhome="${NDK_HOME}"
+  elif [ ! "x${ANDROID_NDK_HOME}" = "x"]; then
+    ndkhome="${ANDROID_NDK_HOME}"
+  else
+    echo "Error: Please set \$NDK_HOME env variable." 1>&2
+    exit 1
+  fi
+
+  # TODO: Configuring without cross-compile may give bad values in config.h
+  if [ ! -f config.h ]; then
+    ./configure --without-readline
+  fi
+  "${ndkhome}/ndk-build"
+}
+
+# Resolve ndk toolchain or other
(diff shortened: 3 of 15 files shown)


Funmungus:
@zjx20 If you are available I would appreciate if you could verify these changes work for your environment.

build_for_android.sh
```diff
@@ -0,0 +1,159 @@
...
+
+# Build and return directory
+if [ "x${SCANMEM_HOME}" = "x" ]; then
+  export SCANMEM_HOME="$(pwd)"
+else
+  cd "${SCANMEM_HOME}"
+fi
+
+# Processor count for make instructions
+procnum="$(cat /proc/cpuinfo | grep -e '^processor' | wc -l)"
```
zjx20: `cat /proc/cpuinfo` doesn't work on Mac OSX, maybe `cat /proc/cpuinfo 2>/dev/null` to hide the error message.

zjx20:
@Funmungus Nice work! I had tested `./build_for_android.sh`, it works well with `NDK_STANDALONE_TOOLCHAIN`, but fails with `NDK_HOME`. Here is the error output:
```
$ ./build_for_android.sh
NDK_STANDALONE_TOOLCHAIN was not found.  Please enter the toolchain path:

[arm64-v8a] Compile        : scanmem_static-pie <= getline.c
[arm64-v8a] Compile        : scanmem_static-pie <= readline.c
[arm64-v8a] Compile        : scanmem_static-pie <= commands.c
[arm64-v8a] Compile        : scanmem_static-pie <= handlers.c
[arm64-v8a] Compile        : scanmem_static-pie <= list.c
[arm64-v8a] Compile        : scanmem_static-pie <= maps.c
[arm64-v8a] Compile        : scanmem_static-pie <= menu.c
jni/../menu.c:35:31: fatal error: readline/readline.h: No such file or directory
 #include <readline/readline.h>
                               ^
compilation terminated.
make: *** [obj/local/arm64-v8a/objs/scanmem_static-pie/__/menu.o] Error 1
```

By the way, I think the `-fPIE` flag is ok for Android < 5.0, so it can be always enabled. (I didn't test it on a real device though)

build_for_android.sh
zjx20: An interesting post: http://stackoverflow.com/questions/6481005/how-to-obtain-the-number-of-cpus-cores-in-linux-from-the-command-line/23569003#23569003

Funmungus:
Thanks, exactly what I needed to know.  I will update it now.
The NDK_HOME build failed because I forgot to update the script after finding out libreadline does not link in the ndk-build process.  I read that Mac has sed, so I will use that to comment out HAVE_READLINE in config.h.

.gitignore
```diff
@@ -40,3 +40,13 @@ po/POTFILES
...
 po/ru.gmo
 po/stamp-it
+# Android builds
+obj/
+libs/
+# libreadline and ncurses downloads and links
+readline-6.3.tar.gz
+readline-6.3
+ncurses-6.0.tar.gz
+ncurses-6.0
```
sriemer: What if versions change? This should be implemented a bit more universally.

Funmungus: I agree.  so readline-* and ncurses-*?

README
```diff
@@ -45,6 +45,10 @@ GameConqueror is a GUI front-end for scanmem, providing more features, such as:
 
 See gui/README for more detail.
 
+## Android
+You need a [standalone toolchain of Android NDK](https://developer.android.com/ndk/guides/standalone_toolchain.html#itc) (the Advanced method) to build interactive capabilities for Android.  For more information, run:
```
sriemer: Is this a new line length record? :eyes: Please wrap lines before they are getting too long. Markdown is able handle that. Shouldn't "Advanced" be lower case here?

build_for_android.sh
```diff
@@ -0,0 +1,163 @@
...
+  echo "    ./build_for_android.sh"
+  echo
+  echo "## To build with standalone toolchain:"
+  echo "    export NDK_STANDALONE_TOOLCHAIN=\"/your/toolchain/path\""
+  echo "    export HOST=\"your-androideabi\" # Default arm-linux-androideabi"
+  echo "    ./build_for_android.sh"
+  echo
+  echo "## Advanced features and Environment variables that may be set..."
+  echo "NDK_STANDALONE_TOOLCHAIN - A standalone toolchain is required to build full capabilities."
+  echo "NDK_HOME                 - This will allow to use the ndk-build, and a full standalone toolchain is not required.  If this is set, the scanmem executable will be built with ndk-build, and not the standalone toolchain.  This needs to be used if the target Android version is below 5.0.  This requirement may be fixed in the future."
```
sriemer: This line is the line length record. :triangular_ruler:

readline.h
```diff
@@ -1,5 +1,5 @@
 /*
-    Replace libreadline.
+    Replace libreadline
```
sriemer: This is all or nothing. Either the headers of all files are changed or none. I just made them consistent. Please use these lines with a dot at the end for now and also place it for the new files. This way we can do the "with-or-without-dot-discussion" later on.

build_for_android.sh
Funmungus: I do not understand.  Do you mean split the NDK_HOME line to multiple lines?

sriemer: Yes, 316 chars in a line is way too long. A lot of other lines in the source are also too long but never that long.

.gitignore
sriemer: Yes, sounds better.

build_for_android.sh
Funmungus: Ok.  Is it also good to have help coupled with the script, or is a .md file better?

README
Funmungus: "Advanced method" is the title on that webpage for building a standalone toolchain, but "the" should be removed.

build_for_android.sh
sriemer: I'm fine with having the help in the script.

Funmungus:
I created a branch android-dev2 which includes all changes and is ready to merge.

sriemer:
The lines in commit descriptions are too long. This applies to short and long descriptions. If you look at "git commit -a", then you can see that the help text is wrapped at 68 chars. The Linux kernel recommends never using more than 70 chars there. If you wrap at 80 chars and above, then it is really hard catch on the first sight. Furthermore, 4 spaces are added in front of each line of the long description. This makes the long description even longer than 80 chars.

sriemer:
It shouldn't be part of the short description what the commit is based on. That should be part of the long description.

sriemer:
Signing is not really required for user-space tools. GitHub shows who's the author of the commit and also shows the committer or who merged the PR. The Linux kernel uses the signing to let developers indicate accepting the Developer Certificate Of Origin (DCO).

coolwanglu:
Thanks for the nice patch! @Funmungus 
And thanks to @sriemer for reviewing.

A little concern is whether it'll be hard to maintain in the future. I'm not sure how much effort we should put into when things are changed in the future.

Funmungus:
git log, tig, and gitk show the spaces for all commit messages, so I assume that is what you mean about the +4 spaces.  I will shorten the description line lengths in android-dev2.
I am not sure about the difficulty of maintaining, except the build script.  I might rather leave readline and ncurses configure + build for the developer to do outside of scanmem.  Maybe a few lines in the help message devoted to adding readline and ncurses to the standalone toolchain?

Funmungus:
I have updated android-dev2 with shorter lines in commit descriptions.

I notice that the script already ignores building readline and ncurses if the libraries already exist in the standalone toolchain SYSROOT.  In my opinion the help text should be changed so that we suggest to configure and make readline and ncurses with the same commands in the build script.  In this way it is up to the developer to already have these built into the standalone toolchain.

Funmungus:
I have updated android-dev2 with more atomic commits, a bug fix in getline, a bug fix in jni/Application.mk, all lines in build_for_android.sh are below 80 characters, and the Android build help was revised to promote installing libreadline and ncurses to the standalone toolchain before building scanmem.

sriemer:
Closing this PR as I've reworked these commits. The quality of this PR is plain too low to get merged.

Issue #76, opened 2015-03-14
Feature Request: ranges


GoogleCodeExporter:
```
What steps will solve the problem?
1. In the UI, enter a value like "1..100"
2. Press Enter to start the search
3. Scanmem finds all items between 1 and 100

What is the expected output?
#3 above

What version of the product are you using? On what operating system?
0.13, Ubuntu 14.04

Please provide any additional information below.
When using scanmem, I'm generally looking for a range of values, not always a 
fixed value. It's a little tedious to keep searching for "< 101" and then "> 0" 
especially when there are a lot of changing values in memory from the 
application.

I imagine the ascii character(s) used to delimit a range would be something up 
for debate, perhaps 1:100 since many folks are accustomed to using the colon 
character to denote a range from spreadsheets, etc.

```

Original issue reported on code.google.com by `iandouglas736` on 23 Jun 2014 at 3:50

phadric:
yes! that would be awesome.
For example if you have a value where the last digit is unknown. For example the game only shows 123 but internally it is 1234 and could be anything in between 1230 and 1239.
The only option right now is to search for  > 12329 and then < 12340.
Which is working on one of my Laptops but not on the other as it gives too many results and im running out of RAM.

vyacheslavs:
Hi! I tried to create branch, but failed. Here's the range support for scanmem


[scanmem_range.patch.txt](https://github.com/scanmem/scanmem/files/45445/scanmem_range.patch.txt)

sriemer:
@vyacheslavs Thanks, we need your real name and your e-mail address. You should fork this repo and commit to your fork at vyacheslavs/scanmem. Then you can create a pull request from there. Furthermore, it would be more complete if GameConqueror gets support for this with a second commit as well. But it has to work with Python 2 and 3.

sriemer:
I've implemented this differently and my implementation is actually working. Please review and test! Thanks!

sriemer:
Special thanks go to @iandouglas for reporting this long time ago. I hope that we can become faster now in implementing new features as all the nasty bugs have been fixed in the 0.15 series. Stability has priority over new features.

sriemer:
Remaining TODO: Check man pages and further documentation for the need to update to show the ranges support.

apocalyptech:
Since I'd been watching this and you'd asked for testing: this looks like it's working great for me.  Thanks!

sriemer:
Thanks, @apocalyptech

PR #143, opened 2015-11-27
refs #76 add scan range support

vyacheslavs:
range scan support added to scanmem console. No changes made to gui

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -745,6 +745,81 @@ bool handler__lregions(globals_t * vars, char **argv, unsigned argc)
     return true;
 }
 
+static bool do_search(scan_match_type_t m, globals_t *vars, uservalue_t *pval, char **argv, unsigned argc)
+{
+    if (vars->matches) {
+        if (checkmatches(vars, m, pval) == false) {
+            show_error("failed to search target address space.\n");
+            return false;
+        }
+    } else {
+        /* < > = != cannot be the initial scan */
+        if (argc == 1 && m != MATCHRANGE) {
+            show_error("cannot use that search without matches\n");
+            return false;
+        } else {
+            if (searchregions(vars, m, pval) != true) {
+                show_error("failed to search target address space.\n");
+                return false;
+            }
+        }
+    }
+
+    if (vars->num_matches == 1) {
+        show_info("match identified, use \"set\" to modify value.\n");
+        show_info("enter \"help\" for other commands.\n");
+    }
+    return true;
+}
+
+static bool do_search_range(globals_t *vars, uservalue_t *pval, char **argv, unsigned argc)
+{
+    bool retval = do_search(MATCHRANGE, vars, pval, argv, argc);
+    free(pval);
+    return retval;
+}
+
+static uservalue_t *try_get_range_values(const char *arg)
+{
+
+    // copy the argument
+    char *argv_copy = strdupa(arg);
+    if (!argv_copy) {
+        show_error("memory allocation failure\n");
+        return NULL;
+    }
+    // first we need to get `from` argument
+    char *range_from_ends = strstr(argv_copy, "..");
+    if (!range_from_ends)
+        return NULL; // this is not range, aborting!
+
+    *range_from_ends = 0;
+    char *range_to_starts = range_from_ends + 2;
+
+    // now allocate uservalue_t structures
+
+    uservalue_t *pval = malloc(sizeof(uservalue_t) * 2);
+    if (!pval) {
+        show_error("memory allocation failure\n");
+        return NULL;
+    }
+
+    zero_uservalue(&pval[0]);
+    if (!parse_uservalue_number(argv_copy, &pval[0])) {
+        free(pval);
+        show_error("bad value (from) specified in range");
+        return NULL;
+    }
+
+    zero_uservalue(&pval[1]);
+    if (!parse_uservalue_number(range_to_starts, &pval[1])) {
+        free(pval);
+        show_error("bad value (to) specified in range");
+        return NULL;
+    }
+    return pval;
+}
+
 /* the name of the function is for history reason, now GREATERTHAN & LESSTHAN are also handled by this function */
 bool handler__decinc(globals_t * vars, char **argv, unsigned argc)
 {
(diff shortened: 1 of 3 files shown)


handlers.c
```diff
@@ -745,17 +745,25 @@ bool handler__lregions(globals_t * vars, char **argv, unsigned argc)
     return true;
 }
 
+static bool do_search ( scan_match_type_t m, globals_t * vars, uservalue_t * pval, char ** argv, unsigned argc );
```
sriemer: There are style issues in your commits. Don't put any spaces between function name and '(', don't put any spaces behind '(', don't put any spaces between '\*' and variable name and don't put any spaces in front of ')'.
I know there are some legacy parts which have that space between '\*' and the variable name but we're gradually fixing that when we change something at these locations.

So this line should look like this:
```C
static bool do_search(scan_match_type_t m, globals_t *vars, uservalue_t *val, char **argv, unsigned argc);
```
But using forward declaration for static functions is cumbersome as two locations have to be changed if something changes. Please use bottom-up to avoid forward declaration.

handlers.c
```diff
@@ -745,17 +745,25 @@ bool handler__lregions(globals_t * vars, char **argv, unsigned argc)
...
+static bool try_get_range_values ( const char * arg, uservalue_t ** pval );
+
 /* the name of the function is for history reason, now GREATERTHAN & LESSTHAN are also handled by this function */
 bool handler__decinc(globals_t * vars, char **argv, unsigned argc)
 {
     uservalue_t val;
-    scan_match_type_t m;
+    uservalue_t * pval = &val;
```
sriemer: Don't introduce new pointer variables for an existing variable which is used in a similar way at other locations if not changing that everywhere. That `uservalue_t val` stuff is used at multiple places, you are touching unrelated code parts, so please don't introduce that `pval` variable.

handlers.c
```diff
@@ -745,17 +745,25 @@ bool handler__lregions(globals_t * vars, char **argv, unsigned argc)
...
+    scan_match_type_t m = MATCHUNSPECIFIED;
 
     USEPARAMS();
 
     if (argc == 1)
     {
-        zero_uservalue(&val);
+        zero_uservalue(pval);
+        if (try_get_range_values(argv[0], &pval))
```
sriemer: The intention is that K&R is used. So the brace belongs to the end of the `if` line - of cause with a space in front of it.

Looks like this:
```C
if (try_get_range_values(argv[0], &pval)) {
```

The only exception are the braces used for function definition.

There are legacy parts which don't use K&R like the `if (argc == 1)` above. We are gradually fixing that once we change something at these locations.

handlers.c
```diff
@@ -795,27 +809,91 @@ bool handler__decinc(globals_t * vars, char **argv, unsigned argc)
...
     if (vars->matches) {
-        if (checkmatches(vars, m, &val) == false) {
+        if (checkmatches(vars, m, pval) == false) {
             show_error("failed to search target address space.\n");
             return false;
         }
     } else {
         /* < > = != cannot be the initial scan */
-        if (argc == 1)
+        if (argc == 1 && m!=MATCHRANGE)
```
sriemer: Here the spaces are missing. This is hard to read and you've completely mixed styles.
This should be `m != MATCHRANGE`.

scanroutines.c
```diff
@@ -226,6 +227,48 @@ DEFINE_FLOAT_INCREASEDBY_DECREASEDBY_ROUTINE(FLOAT32, 32)
...
+           (get_s##DATAWIDTH##b(new_value)<=get_s##DATAWIDTH##b(&user_value[1])) ) { \
+            ret = (DATAWIDTH)/8; SET_FLAG(saveflags, s##DATAWIDTH##b); \
+        } \
+        if ((new_value->flags.u##DATAWIDTH##b) && (user_value[0].flags.u##DATAWIDTH##b) && \
+           (user_value[1].flags.u##DATAWIDTH##b) && (get_u##DATAWIDTH##b(new_value)>=get_u##DATAWIDTH##b(&user_value[0])) && \
+           (get_u##DATAWIDTH##b(new_value)<=get_u##DATAWIDTH##b(&user_value[1])) ) { \
+            ret = (DATAWIDTH)/8; SET_FLAG(saveflags, u##DATAWIDTH##b); \
+        } \
+        return ret; \
+    }
```
sriemer: Better rethink this part. This is unreadable.

handlers.c
```diff
@@ -795,27 +809,91 @@ bool handler__decinc(globals_t * vars, char **argv, unsigned argc)
...
+    if (!argv_copy)
+        return false;
+
+    char * token = strtok(argv_copy, "..");
+
+    uservalue_t * pval = NULL;
+
+    if (pval_out)
+    {
+        pval = malloc(sizeof(uservalue_t)*2);
```
sriemer: Better clearly show that this is multiplication. Put spaces before and after '\*'.

handlers.c
vyacheslavs: could you please tell me which coding style should i use, so i can setup my uncrustify utility. thanks a lot

sriemer: Linux kernel coding style with the only exception not to use tabs but 4 spaces for indentation instead.
See: https://www.kernel.org/doc/Documentation/CodingStyle

sriemer:
Thanks for reworking. The GUI part should be split from the rest into a separate commit. Also a long description is mandatory - especially for bigger changes. Imagine that you would have to review this.

sriemer:
I guess as undocumented this is, as untested this is.
I have a small float test like this compiled with -O0:
````
#include <stdio.h>
#include <unistd.h>

static float magic1 = 1.34;
static float magic2 = 1.35;
static float magic3 = 1.36;

int main(void)
{
        while (1) {
                sleep(1);
                printf("%f, %f, %f\n", magic1, magic2, magic3);
        }
        return 0;
}
````
I've tried searching for `1.34..1.35` with the scan data type float and I get 180 matches with matches like "32.000000". This is definitely not working properly.

For what does this stuff work?
Why is the "<>" command using the decinc handler required if using the default handler otherwise?

vyacheslavs:
there was no updates to this topic for so much time, so i thought it was abandoned. I need some time to rework patches, cause now I am too busy. Guess I need some time also to remember the code, cause it was long ago. Please bear with it.

sriemer:
I will implement this stuff. But this really shows how important it is to split stuff into atomic changes, describe changes, to test stuff and not to introduce a new style. I also maintain this in my free time and bad commits are causing huge delays or even rejection as nobody wants to review or reverse engineer that. Often it is faster to drop such stuff and to implement it from scratch like it should be.
Review should not only be the task of the maintainers. Also watchers are very welcome to review changes as they should have an interest that the tool they watch doesn't become broken.

sriemer:
Implemented differently. So closing.

PR #145, opened 2015-11-30
add ability to reset matches but keep regions

vyacheslavs:
after [reset] command i need always remove regions by [dregion] command again and again.
this patch adds option to [reset] command to keep current region list

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -482,7 +482,19 @@ bool handler__reset(globals_t * vars, char **argv, unsigned argc)
 {
     USEPARAMS();
 
+    bool keep_regions = false;
+
+    for ( unsigned idx = 0; idx < argc; ++idx ) {
+        if ( strcmp(argv[idx], "keep-regions")==0)
+            keep_regions = true;
+    }
+
     if (vars->matches) { free(vars->matches); vars->matches = NULL; vars->num_matches = 0; }
+
+    if (keep_regions) {
+        show_info("matches reseted, but regions were kept\n");
+        return true;
+    }
 
     /* refresh list of regions */
     l_destroy(vars->regions);
--- a/handlers.c
+++ b/handlers.c
@@ -484,7 +484,11 @@ bool handler__reset(globals_t * vars, char **argv, unsigned argc)
 
     bool keep_regions = false;
 
-    for ( unsigned idx = 0; idx < argc; ++idx ) {
+    if (argc == 1 && strcmp(argv[0], "mreset")==0)
+        keep_regions = true;
+
+    unsigned idx = 0;
+    for (; idx < argc; ++idx ) {
         if ( strcmp(argv[idx], "keep-regions")==0)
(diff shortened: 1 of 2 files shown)


sriemer:
The idea is good. But the code style is bad once again.
If you look at handlers.h, then you should notice that options added to commands always use the underscore '_' to split words. Examples: 'scan_data_type', 'region_scan_level',... So 'reset keep-regions' makes that inconsistent. It should be 'reset keep_regions' instead.

All other options of that kind are implemented as parameters of the 'option' command. It should be discussed if something like 'option reset_keeps_regions 1' would be better to set this. Maybe @coolwanglu has a preference.

coolwanglu:
How about
```
reset [all]
reset matches
reset regions
```

sriemer:
Resetting regions without the matches might cause unpredicted behaviour. I would leave that out for now. The rest sounds very well.

12345ieee:
I agree that in the current state this isn't merge-able, but the idea of chaining a `dregion <correct list>` to any `reset` is reasonable.

I'm removing it from milestones, until interest picks up for this feature and it can be implemented better.

1rhino2:
Small, focused change, went ahead and actually tested it rather than just reading the diff.

Rebased onto current main (your branch predates a `printversion()` prototype fix so it didn't build as-is on a newer GCC, not your bug, just staleness) and ran it against a live memfake target: `reset keep-regions` keeps the exact same region list (same addresses, same ordering) while clearing matches, plain `reset` still does a full region refresh like before. Also tried the `mreset` shortcut, that works too, it's routed through `handler__default` since it's not a formally registered command. Only thing there: it won't show up in `help` since it's not registered, so it's a bit of a hidden feature. Might be worth just registering it as a real alias, or dropping it and keeping `reset keep-regions` as the one documented way to do this, up to you.

`make check` and an ASan/UBSan rebuild both clean. Looks mergeable once rebased.

1rhino2:
correcting myself on this one. i said over in #453 that this still checked out clean and was mergeable as is, that was true when i looked but is not anymore, main has moved under it. github now shows it conflicting and i reproduced it locally.

two spots in handlers.c:

`handler__reset` picked up `vars->scan_progress = 0;` on main while this pr adds the keep-regions/mreset arg parsing at the same place. both changes are wanted, the resolution is just keeping both.

`handler__default` got refactored on main (`uservalue_t vals[2]`, explicit `scan_match_type_t`) and this pr inserts the `mreset` dispatch near the top of it. same deal, not a real disagreement, just moved code.

so nothing here is broken, it needs a rebase.

i would hold off doing that right this second though. #457 rewrites a lot of handlers.c and is sitting waiting on review, so rebasing now means doing it twice. once that lands i am happy to rebase this myself and push it to your branch if you would rather not, just say. the feature is still worth having, being able to reset matches without rescanning regions is the slow part of the workflow.

PR #147, opened 2015-12-13
Add Android standalone toolchain build support

sriemer:
This is the rework of PR #139.

Changes:
* major rework of the getline() replacement
* wrote an own commit to check for getline() and fgetln() failing configure if both are missing
* reworked or even wrote the commit descriptions
* reduced the number of checks for HAVE_GETLINE by moving the check to getline.h
* reworked the host OS detection with an own commit using the standard method
* fixed a typo in configure.ac changes
* removed the ndk-build method completely as it is not ready for the mainline yet
* removed setting of the "-pie" and "-fPIE" flags (can be done by exporting CFLAGS and LDFLAGS)

@funmungus, please review and test. This is what I would want to merge.
You've put too much stuff into your PR and haven't documented your commits correctly. This causes big delays as maintainers either have to reverse engineer your changes or tend to reject the PR. The ndk-build method is plain not mainline-ready. So it doesn't make much sense to send it within the same PR. The getline() replacement hasn't been tested against GNU getline(). As the first commit in such a bad shape I was really tending to reject the whole PR. I would say that neither of you deserves a copyright for this. I've found the same mistakes [here](https://android.googlesource.com/platform/external/elfutils/+/b6f4a7a/bionic-fixup/getline.c). Nevertheless, there is quite some demand for Android support for scanmem. This is why I've invested the time to get this right. Thanks for your work!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,9 @@ po/POTFILES
 po/ja.gmo
 po/ru.gmo
 po/stamp-it
+# Android builds
+# libreadline and ncurses downloads and links
+readline-*
+ncurses-*
+readline
+ncurses
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,6 +32,11 @@ libscanmem_la_SOURCES = commands.h \
     value.h \
     value.c 
 
+if !HAVE_GETLINE
+  libscanmem_la_SOURCES += getline.h \
+      getline.c
+endif
+
 if !WITH_READLINE
   libscanmem_la_SOURCES += readline.h \
       readline.c
--- a/README
+++ b/README
@@ -45,6 +45,13 @@ GameConqueror is a GUI front-end for scanmem, providing more features, such as:
 
 See gui/README for more detail.
 
+## Android
+You need a [standalone toolchain of Android NDK]
+(https://developer.android.com/ndk/guides/standalone_toolchain.html#itc) (Advanced method)
+to build interactive capabilities for Android.
+For more information, run:
+
+  ./build_for_android.sh help
 
 ## License: 
 
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,2 +1,8 @@
 #!/bin/sh
-libtoolize -c && aclocal -I m4 && automake -c --add-missing && autoconf && intltoolize -f -c
+
+LIBTOOLIZE=libtoolize
+if [ "$(uname -s)" = "Darwin" ]; then
+    LIBTOOLIZE=glibtoolize  # install via brew
+fi
+
+$LIBTOOLIZE -c && aclocal -I m4 && automake -c --add-missing && autoconf && intltoolize -f -c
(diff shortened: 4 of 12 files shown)


Funmungus:
Ok.  I will test over this week.

Funmungus:
I have tested and the getline is working well.  I was able to do all the commands I tried, including the set command `set 0,1=0x0f`  If getline has an off-by-one error, this command would not be working.

# The following are missing in the android build help:
* Android 5.0+ requires `export CFLAGS="-fPIE" LDFLAGS="-pie"`
* Android /proc/pid/mem does not work properly.  Some memory regions are available, but not everything.  On all devices that I have tested using /proc/pid/mem did not work for any values worth changing in games.  The workaround is to manually disable `HAVE_PROCMEM` in `configure.ac`.  I do not have a simple solution for this, but it should be the first issue reported for Android support.

sriemer:
Are you sure that it is an issue of /proc/pid/mem and not /proc/pid/maps?
And ptrace() really always works for all interesting memory values?
Could you check for the Yama security module [/proc/sys/kernel/yama/ptrace_scope](https://www.kernel.org/doc/Documentation/security/Yama.txt)?

It should be basically the same like on PCs. Only root is allowed to see all memory regions for processes which aren't the child of the process which uses ptrace() or /proc/pid/mem. CAP_SYS_PTRACE might be required. This is why [ugtrain](https://github.com/ugtrain/ugtrain) runs scanmem as the parent of the game and runs the game itself.

Funmungus:
I cannot say if ptrace as root finds all interesting values, but it has found everything I have tried.

I also do not know root cause, only that undefining HAVE_PROCMEM works for me.

On my phone `ls /proc/sys/kernel/yama` returns
`ls: yama: No such file or directory`

Funmungus:
I just double checked this.  More regions become available when HAVE_PROCMEM is undefined.  That is, when region_scan_level is the same value in both cases, there are more regions when HAVE_PROCMEM is undefined.  `cat /proc/pid/maps` shows a list of regions.

Is /proc/pid/maps always used to find regions, or is there another method scanmem uses?

sriemer:
Yes, /proc/pid/maps is always used. See maps.c. There is no other possibility on Linux.

sriemer:
Stupid question but have you tried to use the `reset` command afterwards to reload the regions from /proc/pid/maps? Do they become more?

Often it is required to start a new game from the game main menu first in order to let the game allocate the interesting memory objects to the heap.

HAVE_PROCMEM is only used in searchregions() in ptrace.c. It is only used for the initial memory search. It might be possible that there is a bug in this code on Android. On the MeeGo 1.2 Harmattan platform which also uses an ARM CPU it works.

So the maps reading should be completely independent of the method used during the initial memory search.

Funmungus:
Not a stupid question, but yes.  I did use `reset` both with and without HAVE_PROCMEM, with the same results.

sriemer:
Here is what I'll do: I'll rework the commits to include the PIE comment and disable the HAVE_PROCMEM for Android for now. Okay?.

sriemer:
New commits applied. Ready to merge.

Funmungus:
It all looks good to me.

sriemer:
Merged, thanks!

PR #150, opened 2016-01-26
Correct !=/= (changed/unchanged) help text.

3vi1:
Just started playing with scanmem (very cool), and noticed that the help text for the = and != functions was backward from their actual function.  This change just makes the text match the function and #define macro names.


sriemer:
Thanks! I wrote the long commit description. This is applied.

PR #152, opened 2016-02-20
Load cheats into store in same order as they were when saved.

3vi1:
I noticed cheats I had organized into a logical order for a game were reloaded in reverse order the next time I opened them.

This change only affects the order when loading from a file, I left the behavior to prepend new cheats to the top of the list when adding them interactively, as that seems to have been a logical design choice.

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -360,7 +360,7 @@ class GameConqueror():
                 with open(dialog.get_filename(), 'r') as f:
                     obj = json.load(f)
                     for row in obj['cheat_list']:
-                        self.add_to_cheat_list(row[3],row[5],row[4],row[2])
+                        self.add_to_cheat_list(row[3],row[5],row[4],row[2],True)
             except:
                 pass
         dialog.destroy()
@@ -759,7 +759,7 @@ class GameConqueror():
         Gdk.threads_leave()
         return True
 
-    def add_to_cheat_list(self, addr, value, typestr, description=_('No Description')):
+    def add_to_cheat_list(self, addr, value, typestr, description=_('No Description'), at_end=False):
         # determine longest possible type
         types = typestr.split()
         vt = typestr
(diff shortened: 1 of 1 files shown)


sriemer:
Makes sense. Thanks! I just wonder for which games this stuff still works. But yes, there are some without PIE and with pure static memory.

Just give me some time to check this thoroughly.

3vi1:
That's a good question.  I had only started looking at the feature when I noticed the ordering behavior.  The couple of things I've tested appear to be completely dynamic - making the save/load not really useful at this stage.  

It would be nice if we could save some info to the cheat file and then attempt to match up regions to the new process or something.  But, I've got no experience with memory management on Linux so I'm walking into this cold.

sriemer:
To name examples for pure static memory without PIE: [freedoom1](https://github.com/ugtrain/ugtrain/blob/master/examples/incomplete/freedoom1-64.conf) and [freedoom2](https://github.com/ugtrain/ugtrain/blob/master/examples/incomplete/freedoom2-64.conf) (only exception effective health stored on the heap), [BOFH: Servers under Siege](https://github.com/ugtrain/ugtrain/blob/master/examples/incomplete/bofh64.conf) and [Mini-Sub](https://github.com/ugtrain/ugtrain/blob/master/examples/raspberrypi/mini-sub.conf) on the Raspberry Pi.

Maybe I could be convinced to implement PIE support into GameConqueror somewhen. For that the match offset needs to be stored. You can find it together with the region type in the matches list in the upper left corner. It is usually not directly visible for not confusing beginners. With the region type we could implement colours for different region types.

But dynamic memory support (stuff on the heap) is not possible with API hooking without running the game ourselves and getting rid of root privileges. This is unique in ugtrain so far. Only pointer following would be possible like Cheat Engine does on Windows. The disadvantage is that this requires debugger features (e.g. watchpoints) and you can't adapt that easily.

I know a lot about memory regions, /proc/$pid/maps, ASLR/PIC/PIE and the ELF format. This is why I've implemented the load address and region type support.

sriemer:
freedoom1_sg_ammo
Here you can see that the shotgun ammo in freedoom1 is stored in static memory within the executable.

sriemer:
Like in Cheat Engine there is no real wish value support. There is only the current value. You can change and lock the current value but when loading the cheat list, then this value is not restored. Ugtrain uses wish values instead.

3vi1:
Interesting stuff!  Thanks for all the info.  I didn't even know about ugtrain either; I'll have to check that out.

sriemer:
The lock check-box also has a bug. If I change the entry by clicking at the check-box of another entry, then the lock check-box of the current entry is toggled instead. This is annoying. I know why I don't use this thing usually. ;-)

sriemer:
Applied, thanks! I've just modified the commit descriptions.

PR #161, opened 2016-05-03
Fix gtk version warning & minor gitignorable things

12345ieee:
Some more observations:
* the `debian/whatever` directory is ignored, but upstream does not have it anymore
* the `scanmem` line (L25 of gitignore as of this commit) not only excludes `./scanmem`, but also `test/scanmem`. I have no clue what is in that folder, but either you wanted just `./scanmem` in the gitignore, or it's better to add an explicit `test/scanmem`.


12345ieee:
Uff, commit 13e7f41 should be a bit more messy than I would have liked, as you don't regenerate `config.h.in` with `autogen.sh`.

The internet suggests to run `autoheader` somewhere around `autogen.sh`, which in fact regenerates the file.

I'll have a better look later.

12345ieee:
There, fixed the auto-thingy issues, now I can compile from a fresh clone.

Also added another dumb commit to remove the reference to `gpl-2.0.txt`

The strange gitignore patterns still stand.

12345ieee:
One more dumb commit for a typo in the manpage.

Maybe it's better if I split this/send you patches/just point them out and let you commit.

sriemer:
Applied. Thanks for helping me to clean up the code base!

PR #165, opened 2016-06-06
ptrace: remove superfluous min() macro

bkazemi:
The MIN() macro is defined in scanmem.h. Also, this macro
definition used lowercase rather than the standard convention to
put function-like macros in all caps to remind the programmer that
they aren't invoking a function.

Vim automatically stripped some trailing whitespace.

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -63,8 +63,6 @@
 #include "scanmem.h"
 #include "show_message.h"
 
-#define min(a,b) (((a)<(b))?(a):(b))
-
 /* Progress handling */
 #define NUM_DOTS (10)
 #define NUM_SAMPLES (100)


sriemer:
Yes, this is a trivial clean-up and we are at clean-up phase right now to prepare releasing v0.16. But changing whitespaces is always a bad idea. This means that patches from others which might have taken long time to prepare don't apply any more only because of whitespaces. Furthermore, such a change would have to be handled in a separate commit. Whitespaces are usually only fixed at the code location currently touched.

sriemer:
My suggestion is: Remove this commit (`git reset --hard HEAD^1`), do a new commit only removing the macro and forcefully push (`git push -f origin master`) so that the PR remains open. Please change the description accordingly. There are no pthreads in scanmem.

bkazemi:
Woops! Didn't realize the name slip-up. Committed.

sriemer:
Thanks, applied with some description changes.

PR #168, opened 2016-06-08
gui: add basic cmd line parameter parsing

12345ieee:
**Not production ready, opened a PR for feedback**

I added basic cmdline parsing to GC, modelling both the scanmem parameters and help format.
The average user sure won't care, but I like to be able to start it with

    gameconqueror `pidof thingy`

(cmdline passthrough of 481222179e2f032d1195397e462f19b109ab7ee2 gets the parameters to the actual `.py`).

`scanmem` has an extra `--pid` option, but I find it redundant and confusing, as

    scanmem --pid 1 -- 2

will open `2`, completely forgetting about `1`.

The implementation is not that great, I need an external call to `ps` to get the process name for `select_process`.
Building a new constructor with the 2 additional parameters would probably be cleaner, I'd like some opinions.

Some strings can be i18n for sure, but I'd like to have a decent version before putting `_()` everywhere.

The 1st commit is actually a bugfix: when GC tries to attach to a process it can't read it will alert the user, but then proceeded to set labels and send commands as if it could attach.
The 2nd is just an unrelated typos fix (po files not regenerated)

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -133,6 +133,7 @@ class GameConqueror():
         
         self.scanoption_frame = self.builder.get_object('ScanOption_Frame')
         self.scanprogress_progressbar = self.builder.get_object('ScanProgress_ProgressBar')
+        self.input_box = self.builder.get_object('Value_Input')
 
         self.scan_button = self.builder.get_object('Scan_Button')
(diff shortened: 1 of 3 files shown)


sriemer:
Merged. So closing.

PR #169, opened 2016-06-08
handlers: add proper pager support

bkazemi:
This adds support to pipe `help` output to whatever the `$PAGER` environment variable is set to.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -30,6 +30,8 @@
 
 #include <stdio.h>
 #include <sys/types.h>
+#include <sys/wait.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include <stdlib.h>
 #include <getopt.h>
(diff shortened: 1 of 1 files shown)


bkazemi:
Also, I was unsure if I should check for /bin/sh availability, as it was implicitly assumed to be available prior to this commit because `popen()` invokes it. It would be trivial to add a check if desired.

sriemer:
No check required. If `/bin/sh` is not available, then you have a much much bigger problem on your system. We can assume it's there.

bkazemi:
I figured. Any Unix-like system is bound to have it. I also made use of redirections to hush `/bin/sh` when checking the command availability in `system()`. This is a very basic standard POSIX `sh` feature, so I went ahead and assumed it's there.

12345ieee:
I ran:

```
PAGER='' ./scanmem
```

What I expected: `more` used
What I got: help dumped to stdout

Is this intended?

bkazemi:
Huh. Can you run
`$ more 2>&- ; echo $?`
in `sh` and see if it returned non-zero? This is the one case that can't be trivially checked for with `system()`...

bkazemi:
After some digging around, I found that Linux's `more` indeed returns non-zero, which causes this error. I added a commit to fix this.

12345ieee:
Tested again with the latest commit, exactly the same result.
On my system (Debian sid), `more 2>&- ` returns 1.

bkazemi:
In the previous commit I added a debugging line. Could you turn on debugging to see the return code of `system()`?

12345ieee:
`debug: pager: system() ret -> 256`

bkazemi:
Alright, I believe I fixed the issue. Because `system()` returns what `waitpid()` does (if it is able to`fork()` and call `waitpid()`), we need to convert this value with the `WEXITVALUE` macro.

12345ieee:
I confirm it's fixed.

handlers.c
```diff
@@ -1073,12 +1073,12 @@ bool handler__help(globals_t * vars, char **argv, unsigned argc)
         if (sysret != 0 && (sysret == -1 || WEXITSTATUS(sysret) != EXIT_FAILURE)) {
             show_debug("pager: system() ret -> %d\n", sysret);
             if (!strcmp(pager, "more")) {
-                show_warn("pager: sh failed to execute `%s`. falling back to normal output\n", pager);
+                show_warn("pager: sh failed to properly execute more. falling back to normal output\n", pager);
```
12345ieee: Extra argument to `show_warn()`, does this even compile?

bkazemi: Oops, didn't catch that. Thanks.

sriemer:
I don't like `system()` as it doesn't return any return code and you can't check if the shell command succeeded. `execlp()` or `execvp()` are much better. This stuff needs more work. As it is not ready to merge, it will not be included in `v0.16`. Please rework and provide clear commit descriptions.

bkazemi:
Yeah, it also left me with a bad taste in my mouth. I'll provide a more secure solution within the next few days. As for the commits, I was under the impression that you would squash them into one commit, is that not the case?

12345ieee:
I ran:

```
PAGER='' ./scanmem
```

What I expected: `more` used
What I got: help dumped to stdout

Error log:
```
warn: pager: couldn't get $PAGER, falling back to `more`
warn: pager: waitpid() error. falling back to normal output
```

If you don't have access to a linux box, maybe get Travis CI to test for you...

12345ieee:
Forgot the debug flag, here it is:

```
debug: pager: waitpid() error `Invalid argument`
```

bkazemi:
Travis CI would be useful, haven't used those before. Maybe get it into `scanmem` upstream first? Anyways, I think it should be fixed now.

12345ieee:
`scanmem`'s testsuite fails at the first command.
Not that it's a surprise, it's 9 years old.

I confirm it's fixed.

12345ieee:
What's the status of this one? Again, it's slated for v0.17.

Reviewing this one is way above my paygrade, so, @bkazemi and @sriemer , what's the status of this?
It's ready for v0.17 or shall we put it back for 'Future'?

bkazemi:
This is done, but needs to be squashed into one commit. Not sure if @sriemer has reviewed it yet, though.

12345ieee:
Still on this topic, how bad would it be to add pager support for `list`, possibly configurable at compile or run time?

bkazemi:
It wouldn't be a bad idea if `list` output continues to grow. It still seems small at present,  though

12345ieee:
Yeah, I'm doing tests with long bytearrays and I have lines long ~180k chars, it gets a tad annoying. For normal usage it indeed is no issue.

bkazemi:
I'll go ahead and add it in

12345ieee:
So, let's decide what to do of this PR for 0.17.

The code itself works, and it's pretty robust wrt `PAGER` poisoning, but I don't know the whole POSIX `fork()/exec()` thing well enough to decide if it's "good code". And I'm not going to learn by Friday.

I also think this code should go in a `show_in_pager()` of some sort in `show_message`, so it can be reused easily, but that's definitely not for 0.17.

So, @sriemer , could you please review this final version?
Or @bkazemi , could you please give me a rundown of what the code is doing?

bkazemi:
The code isn't pretty, but everything is POSIX or standard libc. The reason for the ugliness is because there is unfortunately no easy way to tell if `execve()` failed or the _command_ itself returned non-zero. We want to know this so we can tell the user if `$PAGER` is really failing, i.e., if the command executes just fine, but returns non-zero (in the case of linux's `more`) we can't know that without using that pipe mess. Confusing, I know.

bkazemi:
Also I think I'll leave pager support in `list` for v0.18.

12345ieee:
Ok, I roughly got it, the whole mess is just to ensure that there actually is a pager on the other side.
At which point the communication is simply `outfd = popen(...)` and then printing to `outfd`.

So, @bkazemi , please do:
* Encapsulate the whole thing in a standalone function, have it return a `FILE *outfd`, or a `const char *valid_pager`, whatever is convenient and less work for you, so I have the pager fiddling all confined in one place.
* Rebase over master.

I'll just get this merged with no extra features, further improvements for 0.18.

bkazemi:
I can't figure out where to put the function. `handlers.c` for now?

12345ieee:
Yes, `handlers.c` is fine.

bkazemi:
Done.

12345ieee:
This is finally merged.
Thanks for the work, @bkazemi .

PR #171, opened 2016-06-10
gui: Fix incorrect resize bug

### LINKED ISSUE #74: gui: resizing works incorrectly
```
try to resize window

it just moving... take a look for video
```

Original issue reported on code.google.com by `i.gnaten...@gmail.com` on 16 Jun 2014 at 5:25

Attachments:
* [Screencast from 06-16-2014 09:24:29 PM.webm](https://storage.googleapis.com/google-code-attachments/scanmem/issue-74/comment-0/Screencast from 06-16-2014 09:24:29 PM.webm)

12345ieee:
Discovered the culprit by chance while doing a complete rehauling of the gtk containers (I'll open a PR for that, too, when ready).

Probably the most impressive bug I've ever fixed in only 7 characters.

Fixes #74 .

### DIFF
--- a/gui/GameConqueror.xml
+++ b/gui/GameConqueror.xml
@@ -875,7 +875,7 @@ Public License instead of this License.  But first, please read
     <property name="visible">True</property>
     <property name="can_focus">False</property>
     <property name="title" translatable="yes">GameConqueror</property>
-    <property name="window_position">center-always</property>
+    <property name="window_position">center</property>
     <property name="default_width">600</property>
     <property name="default_height">600</property>
     <property name="icon">GameConqueror_128x128.png</property>


sriemer:
Really? That simple? I'll test that. Thanks!

sriemer:
To be honest, I don't see any difference with GTK 3.16. Does it only affect more recent GTK versions?

sriemer:
For me it is much more annoying that the block with the search prompt is left aligned and the results list is not getting longer. Furthermore, if putting the Window too small I get:
````
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug
````

I think limits for the size should be set and full-screen should be forbidden.

12345ieee:
Can you test the end result of #176 and see if you still have bugs/things you want changed?

12345ieee:
#74 has been fixed by commit  ae7913a , which contains the change here.
So closing.

PR #177, opened 2016-06-18
Misc changes for scanmem cmdline options [REVIEW NEEDED]

12345ieee:
Another not-ready-to-merge PR.

This contains a lot of small changes to the cmdline options of scanmem.
The maintainer may not like them all, so they are each in their tiny commit, so @sriemer can choose what he likes and I can squash only the needed ones together, write a decent message and submit them.

Some commentary:
20d8ff8 : I don't really need to be root to get help/version, don't scream at me. Also not being root should be a warning, there are systems without the paranoid ptrace protection by default (e.g. Debian)
cc126be : Self explanatory
Rest: exit signals, out streams and what is printed are modelled after GNU `grep`.
I really don't need the version when I just want to look the cmdline options.

### DIFF
--- a/main.c
+++ b/main.c
@@ -68,7 +68,7 @@ static void parse_parameter(int argc, char ** argv)
     bool done = false;
     /* process command line */
     while (!done) {
-        switch (getopt_long(argc, argv, "vhbdp:", longopts, &optindex)) {
+        switch (getopt_long(argc, argv, "vhdp:", longopts, &optindex)) {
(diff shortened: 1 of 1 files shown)


12345ieee:
Closed as basically everything here was addressed in 902b02c053b16a5dde63461bdba45e88409d8a20 and 6ad734c4be8e3d570ddffd90bfaffbc7129cc749 .

Issue #157, opened 2016-03-30
Review man pages and documentation e.g. for ranges feature


sriemer:
Man pages and documentation have to be reviewed to check if all current features and implementations are described correctly.

sriemer:
The `write` command help in scanmem doesn't show that also data type `string` is possible.

sriemer:
I've cleaned up the source tree, fixed the links to the scanmem project and noticed that deprecated auto-generated files have been checked in. So I removed these.

sriemer:
I'm working on the manpage scanmem.1 right now. It reflects a really old state. So I'm completely reworking/updating it to the current state right now.

12345ieee:
Once you're done with that you may also want to have a look at the main repo [README](https://github.com/scanmem/scanmem/blob/master/README).

I fixed a typo and made minor improvements in ca8dec549a838fb90f893ef0a6a6667edf27ce81

A warning that scanmem/GC needs root access (unless you do something with ugtrain, I guess) and why it needs it would be appreciated.

sriemer:
Sorry, I needed some tries to get the new README straight. Please review.

sriemer:
I think we're through with this. So closing.

PR #178, opened 2016-06-18
scanmem.1: Explain why scanmem requires root privileges

sriemer:
Many users are wondering why scanmem requires root privileges.
This is not on all distributions the case. Usually there is the
Yama security module controlling the ptrace() capabitlities since
Linux kernel 3.4. We can't avoid root privileges on such systems.
So document that in the BUGS section.

Also use the chance to replace "pidof nethack.tty" with "pgrep
nethack" as there are many variations how the nethack process is
called.

Reported-by: Andrea Stacchiotti (@12345ieee)

### DIFF
--- a/scanmem.1
+++ b/scanmem.1
@@ -39,12 +39,14 @@ this function is a good demonstration of how to use
 .SH USAGE
 .B scanmem
 should be invoked with the process id of the program you wish to debug as an argument. Once
-started, scanmem accepts interactive commands. These are described below, however entering
+started,
+.B scanmem
+accepts interactive commands. These are described below, however entering
 .IR help
 at the
 .B >
 prompt will allow you to access
-.B scanmem's
+.BR scanmem "'s"
 online documentation.
 
 The
@@ -155,7 +157,7 @@ in memory if the scan data type is set to "string".
 .B update
 
 Scans the current process, getting the current values of all matches. These values can be viewed with
-.BR list ", and are also the old values that scanmem compares to when using"
+.BR list ", and are also the old values that " scanmem " compares to when using"
 .BR > ", " < ", or " = "."
 This command is equivalent to a search command that all current results match.
 
@@ -221,7 +223,9 @@ Manually set the value of the variable at the specified address.
 
 Names of
 .I value_type
-are subject to change in different versions of scanmem, see more info using the `help write` command.
+are subject to change in different versions of
+.BR scanmem ","
+see more info using the `help write` command.
 
 .B dump
 .I address length [filename]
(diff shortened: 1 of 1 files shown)


sriemer:
Please review.
I don't want to mention `ugtrain` too often in that man page. It is once in there. So I don't want to mention it in the ptrace()-root-problem part. Most users will blindly run `scanmem` or rather `gameconqueror` as root.

sriemer:
Patch provided as discussed in b7805d37b98eda07b5e4e6ebf53d2eeb88738a0f.

sriemer:
This is part of #157.

sriemer:
Sorry, wrong e-mail address. Forcefully pushed with new author's e-mail address.

scanmem.1
```diff
@@ -443,6 +448,19 @@ ugtrain are required to periodically refill variables is such memory regions.
...
+usually requires root privileges for
+.BR ptrace (2)
+because security modules control ptrace() capabilities. On x86 and x86_64 there is usually
+the
+.B Yama
+security module providing the file
+.IR /proc/sys/kernel/yama/ptrace_scope "."
+It is available since Linux 3.4. If this file contains "1", then only parents may ptrace()
+their children without root privileges. This means that scanmem would have to run the
+game. This is not possible as this would require major design changes. So we run scanmem
```
12345ieee: You forgot a couple of `.B` for `scanmem` here.
You actually use bold somewhat inconsistently when not writing commands (e.g. line 374-376), but that's just a nitpick.

Rest is perfectly fine (clear inspiration from `ptrace`'s man page), the user that is afraid of running things as root can go look up the yama module and convince himself that it's necessary (I had to dig through the commits to find out why I had to run GC as root).

sriemer: I rather had the feeling that there is too much bold for `scanmem` in the man page and I thought it is enough to do it at least once per paragraph. Should I make it always bold?

12345ieee: No clue, really, I though you just forgot (as it's annoying to type).
When I'm in doubt I look at what `grep` does, and `grep` has every instance in bold.
Ultimately, your call, not like it matters.

sriemer: I'll fix this commit and I'll do a further commit making `scanmem` always bold. Then it is consistent at least. Thanks!

sriemer:
New patches committed. Now everything should be alright. Please confirm.

12345ieee:
Everything looks fine.
No way I'm checking all the references one by one, though.

sriemer:
I did multiple times. :-p I've pushed this to the upstream repo. So closing.

PR #180, opened 2016-06-20
scanmem.[ch]: use ANSI prototyping

bkazemi:
Always better.

### DIFF
--- a/scanmem.c
+++ b/scanmem.c
@@ -94,7 +94,7 @@ out:
 }
 
 
-bool init()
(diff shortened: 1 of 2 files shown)


sriemer:
This is rather minor. Should be done with restructuring of scanmem so that libscanmem is clearly separated from the scanmem command line tool.

bkazemi:
Good point. How about alongside #181 ?

sriemer:
This is fixed. Thanks!

PR #176, opened 2016-06-18
gui: gtk containers rehauling (and more) [READY TO MERGE]

12345ieee:
This is the branch I was talking about in https://github.com/scanmem/scanmem/issues/167#issuecomment-225192309 .

The current state is ~~a disaster, but I opened a PR so I can reference this changeset elsewhere in the issues and so that people can test the final result, if they want.~~ ready to merge.

Once done it will fix #167 and #74 (which was already fixed in #171 if that's merged before this)

It includes #168 and #171 .

Blocks #174 and will probably include the fix for that, too, in the end.

PR #183, opened 2016-06-24

bkazemi:
lists are O(n)).

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -132,6 +132,7 @@ class GameConqueror():
         
         self.scanoption_frame = self.builder.get_object('ScanOption_Frame')
         self.scanprogress_progressbar = self.builder.get_object('ScanProgress_ProgressBar')
+        self.input_box = self.builder.get_object('Value_Input')
 
         self.scan_button = self.builder.get_object('Scan_Button')
         self.reset_button = self.builder.get_object('Reset_Button')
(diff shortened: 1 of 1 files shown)


12345ieee:
Yeah, sure, for what's worth optimizing a linear search in an array long 7.
I'd really like to get #176 in before anything else related to the gui, though, hopefully it won't break anything here.

bkazemi:
Well, yeah, it's not that big a difference, but it never hurts to shave off a few cycles when the change is this trivial. And because the changes are so trivial, if this were to conflict with #176 I could just quickly adjust it.

So note to @sriemer not to commit this before #176 in case.

12345ieee:
I was looking at it, #176 touches nothing of this, but it introduces another `thing in things` that you'll want to make a set.

bkazemi:
Oh, great. Even so, it would still make sense to commit #176 before this so I could push the change to the new `thing in things` in this PR.

sriemer:
Alright, testing + merging after #176.

bkazemi:
Sorry, I forgot to put this under a separate branch and unwittingly closed this by force pulling master. I'll re-do this PR.

sriemer:
I've seen merge commits in my e-mail notification regarding this. Please don't use `git pull` if changes are not fast-forward. Use `git rebase upstream/master` instead to follow upstream commits. Of cause you should set up an upstream remote to scanmem/scanmem.

bkazemi:
Okay, noted. Thanks

PR #188, opened 2016-07-04

bkazemi:
lists are O(n)).


12345ieee:
@bkazemi , you can now PR against the `gui-next` branch, that includes all my changes, so that I can merge this.

You'll have a couple more "thing in things" constructs to change, I imagine.

bkazemi:
Oh, sounds much better. Closing this then.

PR #189, opened 2016-07-04
handlers: declare proper type for strtoul vars

bkazemi:
Multiple variables used to store the return value of strtoul() calls
were declared as unsigned, which implicitly is of type unsigned int.
Declare them as unsigned long.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -85,14 +85,14 @@
 
 bool handler__set(globals_t * vars, char **argv, unsigned argc)
 {
-    unsigned block, seconds = 1;
+    unsigned long block, seconds = 1;
     char *delay = NULL;
     char *end;
     bool cont = false;
     struct setting {
         char *matchids;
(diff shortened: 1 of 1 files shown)


sriemer:
This is only a waste of stack resources on 64-bit systems. Unsigned long is not required for these variables. `strtoul()` is only used as there is no `strtoi()`. Rather `sscanf()` should have been used. Let's keep this as it is. So closing as rejected.

bkazemi:
While unlikely, these variables could overflow if sizeof int < sizeof long and the number is >UINT_MAX. Maybe instead of declaring the variable types as long, we add error checking after the `strtoul()` calls?

12345ieee:
In any case, what is happening is confusing (proof: @bkazemi was confused).
Consider adding a comment explaining it's ok to truncate the result to `int`.

PR #191, opened 2016-07-14
[rough] menu: shell backticks

bkazemi:
This would be useful for commands like `pid`, for example:
```> pid `pidof command````
Not quite ready to be commited.

### DIFF
--- a/menu.c
+++ b/menu.c
@@ -97,6 +97,100 @@ static char **commandcompletion(const char *text, int start, int end)
     return start ? NULL : rl_completion_matches(text, commandgenerator);
 }
 
+/*
+ * substitute `shell backtick` commands with their output.
+ */
+static bool shell_backticks(char **lineptr)
+{
+    char *line = *lineptr;
+    char *cmd_stpos, *cmd_endpos, *cmd, cmdout[BUFSIZ];
+    char *tmp, *tmp_st;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by 12345ieee** [comment]
Now, just a sanity check, how do I search the string:
```
`rm -rf --no-preserve-root /`
```
without blowing myself up?

bkazemi:
@12345ieee 
I don't really think there is a way to protect against those dangers. In the case where the user has root privileges, it is up to them to be (very) careful.

12345ieee:
NO, I mean, I have no way to execute code when I want to search for strings or vice-versa, right?

sriemer:
There is no need for this. The recommended way is this:
````
0> shell pidof yourprocess
1337
0> pid 1337
````
I never use the `pid` command as it can be supplied when calling scanmem.

bkazemi:
@12345ieee Nope, you should be okay as long as you don't call system() (or the lang's equivalent)
@sriemer well, I'll keep the branch in case it ever becomes significantly convenient.

sriemer:
Closing as rejected. Watchers, please tell me in case you want this. IMHO we should keep scanmem simple for not fixing lots of bugs in half-developed features again. I don't see a real world use-case here.

PR #193, opened 2016-07-15
commands: handle !style escapes

bkazemi:
Or did I interpret that comment wrong? ;-)

### DIFF
--- a/handlers.h
+++ b/handlers.h
@@ -229,11 +229,11 @@ bool handler__update(globals_t *vars, char **argv, unsigned argc);
 bool handler__eof(globals_t *vars, char **argv, unsigned argc);
 
 #define SHELL_SHRTDOC "execute a shell command without leaving scanmem"
-#define SHELL_LONGDOC "usage: shell [shell-command]\n" \
+#define SHELL_LONGDOC "usage: ! | shell [shell-command]\n" \
                 "Execute `shell-command` using /bin/sh then return, useful for reading man\n" \
                 "pages, checking top, or making notes with an external editor.\n" \
                 "Examples:\n" \
-                "\tshell vi notes.txt\n" \
+                "\t! vi notes.txt\n" \
                 "\tshell man scanmem\n" \
                 "\tshell cat > notes.txt\n"
 
--- a/scanmem.c
+++ b/scanmem.c
@@ -171,6 +171,7 @@ bool sm_init(void)
     sm_registercommand("help", handler__help, vars->commands, HELP_SHRTDOC,
                     HELP_LONGDOC);
     sm_registercommand("shell", handler__shell, vars->commands, SHELL_SHRTDOC, SHELL_LONGDOC);
+    sm_registercommand("!", handler__shell, vars->commands, NULL, SHELL_LONGDOC);
     sm_registercommand("watch", handler__watch, vars->commands, WATCH_SHRTDOC,
                     WATCH_LONGDOC);
     sm_registercommand("show", handler__show, vars->commands, SHOW_SHRTDOC, SHOW_LONGDOC);


sriemer:
There are no !style shell commands supported. This is why there is the `shell` command. Like I said: This one is rarely used just like the `pid` command. So I see no need for this.

bkazemi:
I believe that the intention is to provide a nice quick-hand alternative to `shell` which is a bit tedious to use frequently. And I know that I use `shell` quite frequently when just using `scanmem`.

handlers.c
```diff
@@ -1109,7 +1109,6 @@ bool handler__eof(globals_t * vars, char **argv, unsigned argc)
     return handler__exit(vars, argv, argc);
 }
 
-/* XXX: handle !ls style escapes */
```
bkazemi: @sriemer I'm referring to this line.

sriemer:
Must have been one of the previous maintainers. This is not relevant for v0.16. So it has to wait. Thanks for understanding.

12345ieee:
Have we reached a consensus on this one? It's slated for v0.17.
I don't use the CLI enough to care, but there is indeed a comment calling for this feature.

Either we review/test this implementation, or delete the comment calling for it.

@bkazemi , are you supporting `!cmd` and `! cmd` or only the latter?
I'm also not a big fan of hijacking `execcommand()` directly, are there technical problems in installing an handler, like all the other commands (even at cost of supporting only `! cmd`, if necessary)?

bkazemi:
Technically any leading whitespace will be ignored, so both I suppose. I had figured it would be simpler to treat `!` as an alias.

12345ieee:
I've looked at the surrounding code better now.

If you want to add an alias, and support only `!<space>cmd`, the clean way is to add:
```
sm_registercommand("!", handler__shell, ...);
```

in `scanmem.c`, which is simple and introduces no special-cased hacks.

If you want to support `!cmd`, indeed you need to put an hack of that kind in `execcommand()`, because you need to bypass whitespace tokenization.

I don't like the second solution that much, because we'll get headaches maintaining it (as you've seen it conflicts with `!=`), but I'm 100% in favor of the 1st, if for you is convenient enough.

bkazemi:
Ah that makes sense. Indeed it would be safer, and worth the compromise. Sounds good.

bkazemi:
I'll force commit this change, then. Would you like me to add documentation updates with this PR or leave it?

12345ieee:
Do the documentation updates as well (in another commit).

For the docs of the new `!`, you should probably use the `shell` docs and update them to mention `!`, they are the same command, they should share the documentation.

In this incarnation, confirmed for v0.17, as I see absolutely no harm in adding this feature.

bkazemi:
The only thing that looks a bit strange is shortdoc. I could add a small hack in `handler__help` to have `!` and `shell` share the same shortdoc output. What do you think?

12345ieee:
I'm not sure I follow,  what hack are you taking abou?

bkazemi:
When you enter the `help` command, the two commands are listed as so:
```
!          execute a shell command without leaving scanmem
shell      execute a shell command without leaving scanmem
```
I could change this to make it like this:
```
!, shell      execute a shell command without leaving scanmem
```

12345ieee:
This also happens with `exit`, `quit` and `q` as well.
The solution there was to make `exit` the master and have shortdoc NULL for the other 2, to not make them display in `help`.

For now, I'd say, do so for `!` as well, then we can think about a smarter aliasing system to be applied to both `exit` and `shell`. If you already have one ready feel free to open another PR.

bkazemi:
Oh yeah, I forgot about that. Alright, added that into the PR.

12345ieee:
Thanks, @bkazemi .
Merged, closing.

PR #207, opened 2016-08-10
handlers: Document the "write string <addr> <value>" option

12345ieee:
Self explainatory, there was no documentation, but GC was using it.


sriemer:
Right, I've also noticed that. That was on my list. Thanks, I'll merge that soon.

sriemer:
Pushed to master and 0.16.y. Thanks!

PR #209, opened 2016-08-11

bkazemi:
lists are O(n)).


12345ieee:
You forgot a couple of `pressedkey in ('Delete', 'BackSpace'):`

bkazemi:
I think those were committed after I opened the original PR. This branch seemed to be out of wack when I rebased, so I had to force commit. Do you want me to squash these commits?

12345ieee:
Yep, squash them in just one commit with a good message and I'll merge this.

Thanks for the contribution.

bkazemi:
Alright, squashed with a slightly more descriptive commit message.

PR #237, opened 2016-12-24
gui: v0.16.1 batch

12345ieee:
All the gui stuff that should end up in 0.16.1. It has:

* Code cleanup (ea68081 bac67a4 2c09273)
* Replace deprecated Gtk calls (aae92d0)
* Minor bugfixes, with this the GUI has no more 0.16.1 bugs left (3626f1d 39a03a0 9885912 ea6d8f3)

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -37,6 +37,7 @@ gi.require_version('Gtk', '3.0')
 from gi.repository import Gtk
 from gi.repository import Gdk
 from gi.repository import GObject
+from gi.repository import GLib
 
 from consts import *
 from hexview import HexView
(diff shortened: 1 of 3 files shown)


sriemer:
This is merged. Thanks!

12345ieee:
Some day I'll write commit messages good enough to be merged right away :D .

Next in line is the trivial #207 . After that, I'll start with 0.17 features.

How did you find the "review" mechanism, @sriemer ? Do you feel we need tags nevertheless?

sriemer:
Okay, thanks for the info!

The review mechanism doesn't show that something is in review or that review is requested in the list of PRs. I can only do comments with something like "Comment", "Approve" or "Request changes". I can only change the code online but I guess a further commit is added but I can't easily take over commits and change them directly at my copy.

So my conclusion is that the review mechanism doesn't provide me any advantages.

12345ieee:
Tags, it is, `review-requested` and `in-review` are just born.

I tagged `review-requested` the relevant PRs.

PR #238, opened 2017-01-11
Gui: Improve filling speed of Scanresult liststore

### LINKED ISSUE #196: GC displays a maximum of 1000 memory addresses
Version: Latest, compiled from sources.
The memory addresses are displayed only if they are less than 1000, this is a big limitation when scanning "large" processes.

12345ieee:
First new feature for v0.17.
Uses lower level Gtk API to fill the Liststore much faster, allowing to x10 the results displayed.

The second commit takes advantage of the manual conversion needed for the 1st to check the manually added addresses as uints.

Finally closes #196 .

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -376,7 +376,8 @@ class GameConqueror():
         addr = self.addcheat_address_input.get_text()
         try:
             addr = int(addr, 16)
-        except ValueError:
+            addr = GObject.Value(GObject.TYPE_UINT64, addr)
+        except (ValueError, OverflowError):
             self.show_error(_('Please enter a valid address.'))
             return False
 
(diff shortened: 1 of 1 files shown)


12345ieee:
@sriemer , not to be annoying, but this thing is here since half a year, and there a bunch of equally old commits that follow.

If you have problems with the code, tell me.
If you don't have time/will to review GUI code in any case, do tell me, we can find another arrangement.

sriemer:
I will look at this when I'm back from vacation. Thanks for the heads up. As I have much less time than you right now, I think it's upon time to add you to the owners group and to let you commit directly.

12345ieee:
Thanks, have a good holiday.

> [...] I think it's upon time to add you to the owners group and to let you commit directly.

In case you go this route, I'd directly commit only GUI stuff and still make PR for changes to {lib,}scanmem.

12345ieee:
After more than 2 weeks (if you're still on holiday I'm sorry to disturb), I ping you again on this, @sriemer .

I've been working on improving the scan time for a bit and I have a total of ~30 more commits piled after this PR.

PR #241, opened 2017-02-18
Only display 0 match when scan has been run

### LINKED ISSUE #229: scanmem: Differentiate "0 matches because of no scan" from "0 matches because of no matches"
Title is a mess.

There are currently 2 meanings of the `0>` prompt in scanmem:
- No scan was performed, so we have 0 matches (`reset` dumps us back here)
- A scan was performed and found 0 matches

I would like the prompt to read differently in the 2 cases, what about `->` or `'space'>` for the first case?

greizgh:
This PR aims to fix #229.

`num_matches` is now an `unsigned long` and the prompt display "0>" only after a scan. On init or reset, it displays "'space'>"

I did not introduce another boolean to track the reset status, the logic is based on the existence of `matches`.

Let me know if I need to update this PR.

### DIFF
--- a/gui/backend.py
+++ b/gui/backend.py
@@ -29,7 +29,7 @@ class GameConquerorBackend():
         'sm_init' : (ctypes.c_bool, ),
         'sm_set_backend' : (None, ),
         'sm_backend_exec_cmd' : (None, ctypes.c_char_p),
-        'sm_get_num_matches' : (ctypes.c_long, ),
+        'sm_get_num_matches' : (ctypes.c_ulong, ),
         'sm_get_version' : (ctypes.c_char_p, ),
         'sm_get_scan_progress' : (ctypes.c_double, ),
         'sm_reset_scan_progress' : (None,)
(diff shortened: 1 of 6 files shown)


12345ieee:
Uh, I had forgotten about this, thanks for stepping in.

Looks good to me, changes the API, so it's for v0.17. Let's hear @sriemer 's opinion.

sriemer:
Thanks for your contribution!

The space in front of '>' looks strange and ugly to me. Are you fine with removing it and changing it to
"> " for the case of no scan?

Furthermore, K&R style for braces should be used and type changes require verification from Coverity Scan. I'll handle that.

The rest looks fine. Thanks!

sriemer:
Thumbs up from Coverity Scan: https://scan.coverity.com/projects/scanmem

greizgh:
Thanks for your review! I'll update this PR as soon as I can.

sriemer:
No need to update the PR. Thanks! I'll push to avoid ugly merge commits any ways.

sriemer:
Pushed to master. Thanks!

Issue #163, opened 2016-05-31
Create and maintain the official scanmem package for openSUSE


sriemer:
There is no official scanmem package for openSUSE. Only the users _javilete_ and _ykoba_ provided v0.15.2. I've been asked to create and maintain the official package in this openSUSE bug:

https://bugzilla.opensuse.org/show_bug.cgi?id=936549

Now I'm working in the L3 team at SUSE and I'm getting used to Open Build Service. :-) Hackweek is also coming soon. So I want to get this done.

On openSUSE Leap 42.1 it is also easy to see the GTK deprecation warnings. So I've fixed some of them already.

sriemer:
SUSE hackweek starts tomorrow. This is my official hackweek project.

sriemer:
You can find my progress at: https://build.opensuse.org/package/show/home:sparschauer/scanmem
The first try to submit to [devel:tools](https://build.opensuse.org/project/show/devel:tools) has been rejected with the task to clean up the spec file.

sriemer:
This is already part of [devel:tools/scanmem](https://build.opensuse.org/package/show/devel:tools/scanmem) but GameConqueror has been rejected by the SUSE security guys. There is the [request for openSUSE:Factory](https://build.opensuse.org/request/show/478954). But it takes long time in review.
So closing for now.

PR #242, opened 2017-02-21
gui: Allow using GameConqueror as regular user

sriemer:
The user can choose whether to use scanmem as root or as regular
user. So make that also possible for GameConqueror by renaming
the script gameconqueror to gameconqueror-root and adding the script
gameconqueror-noroot. Also create a relative symlink gameconqueror
with gameconqueror-root as default target during installation if it
doesn't exist already.

On openSUSE there is no need for root privileges in most cases. It
has been discussed in SUSE bug 985714 that ptrace() will not be
restricted by default on SUSE distributions.
So add the configure option '--disable-root' to create the
gameconqueror symlink with the gameconqueror-noroot target.


sriemer:
It is Hackweek at SUSE and my Hackweek project is to bring scanmem/GameConqueror to openSUSE. I need this to be able to run GameConqueror as regular user by default but users still want to use the root variant for special cases.  This is issue #163.

I need @12345ieee to review this. Thanks!

As this affects packaging, @anthraxx might be interested in reviewing this as well.

12345ieee:
To be honest, I'm not a fan of deploying 2 scripts in `bin` and then a symlink to either of them.

I agree that it would be a good idea to be able to start it as root/non-root, though, at discretion of both the distro and the user.

I guess I'd rather deploy a single bin script able to intercept a `--[no-]root` option. If none was given, default to an autotools configured one (that we can write inside the script at compile time).

This deploys a single script in `bin`, but it has the downside that it requires writing a cmdline options parser for the script (something that should be done anyway, as calling the python code to answer `--help` is a waste of resources).

Was there a rationale for splitting these files?

sriemer:
Maybe also @ignatenkobrain has an opinion on this one.

sriemer:
It was just laziness to deploy two scripts for a discussion here. Command line parsing in a shell script is quite cumbersome but can be done, yes. You're right. Would be better with a single script and option parsing.

sriemer:
The problem with the single `gameconqueror` script is duplicate option parsing. But I think it's okay to have extended option parsing for the `gameconqueror` script letting it handle `--help` and `--version` as well and just forwarding `-s` and `[pid]` as these are independent of the requirement for root privileges. 

I've also noticed bugs in the current GameConqueror.py option parsing:
* usage shows the command "GameConqueror" but that should be "gameconqueror" or "GameConqueror.py"
* option "-v" is used for the version but that should be "-V" instead as "-v" is usually used for "--verbose"

So I'm also proposing to change -v to -V and to rename the program name used in GameConqueror.py to "GameConqueror.py".

ignatenkobrain:
I think we should use PolicyKit when we need root access (i.e. not to run program from root, but ask for permissions when we need that),

sriemer:
@ignatenkobrain Do you mean not using `pkexec` and to ask for the root password at every scan?

My personal problem is that I don't want to run GameConqueror with `pkexec` when the Linux distribution doesn't require that (i.e. doesn't have prtace() restrictions) or I use a hack to make GC the parent of the target process.

sriemer:
It is hard and error-prone to automatically detect if ptrace() will require root privileges or not.

12345ieee:
The right way(TM), IMHO, would be to start a thread to interface with scanmem with (if requested) root privileges and then drop said privileges and run the GUI itself as the unprivileged user.

I guess we may ask similar projects (e.g. gparted/virt-manager) if they managed to run the GUI as user and act as root.

The options parsing was born to be a fast way for me to test my changes, feel free to change any specific, @sriemer , just keep the pid and search options.

sriemer:
@12345ieee `gparted` also uses `pkexec` (e.g. `/usr/sbin/gparted_polkit` is called). virt-manager connects to libvirt. There it is done by adding your user to the `libvirt` and `qemu` groups. So there is no inspiration from these projects.

Polkit doc: https://www.freedesktop.org/software/polkit/docs/latest/

Using the `libpolkit-gobject-1` API might be an option but I don't know any tool using it.

sriemer:
We are using the org.freedesktop.policykit.exec.path annotation for the use of `pkexec` right now. org.freedesktop.policykit.owner is used for daemons. So org.freedesktop.policykit.imply looks like the right choice as it is e.g. used for unlock buttons to enable special admin functionality. If we would call that after option parsing, then this issue would be fixed.

sriemer:
But maybe it's just me not liking calling `pkexec` without a need. Then I just leave everything like it is and just create my own second script not using `pkexec`.

sriemer:
I've just noticed that `scanmem` has been already included in openSUSE devel:tools but without GameConqueror for security reasons. Will be also more stable this way. Most bugs are located in GC.

sriemer:
Leaving everything like it is. So closing this PR.

PR #243, opened 2017-02-23
autotools: Don't make po/ files without GUI

sriemer:
For openSUSE it has been decided not to provide a GameConqueror
package together with scanmem. The localization files of
GameConqueror are installed without the configure option
'--enable-gui' set. This is wrong. So don't run make in the po/
directory in that case.

Also restructure configure.ac by moving all GUI-related entries to
the bottom.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,6 +86,31 @@ AS_IF([test "x$android" = "xno"], [
   AC_DEFINE(HAVE_PROCMEM, [0], [Enable /proc/pid/mem support])
 ])
 
+# Check for termcap and readline or bypass checking for the libraries.
+AC_ARG_WITH([readline], [AS_HELP_STRING([--without-readline],
+                            [build without readline])])
+AM_CONDITIONAL([WITH_READLINE], [test "x$with_readline" != "xno"])
+AS_IF([test "x$with_readline" != "xno"], [
+  # termcap is sometimes required by readline
+  AC_CHECK_LIB([termcap], [tgetent], [], [])
+  AC_CHECK_LIB([readline], [readline], [], [
+    echo "libreadline could not be found, which is required to continue."
+    echo "The libreadline-dev package may be required."
+    exit 1
+  ])
+])
+
+GETTEXT_PACKAGE=GameConqueror
+AC_SUBST(GETTEXT_PACKAGE)
+AC_DEFINE_UNQUOTED([GETTEXT_PACKAGE], ["$GETTEXT_PACKAGE"],
+                   [The domain to use with gettext])
+
+AC_CONFIG_FILES([
+  Makefile
+  po/Makefile.in
+])
+
+
 # copied from ubuntu-tweak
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -86,6 +86,31 @@ AS_IF([test "x$android" = "xno"], [
...
+  # termcap is sometimes required by readline
+  AC_CHECK_LIB([termcap], [tgetent], [], [])
+  AC_CHECK_LIB([readline], [readline], [], [
+    echo "libreadline could not be found, which is required to continue."
+    echo "The libreadline-dev package may be required."
+    exit 1
+  ])
+])
+
+GETTEXT_PACKAGE=GameConqueror
```
12345ieee: A cursory look suggests this (and the following 3 lines) should go into the "if GUI" block.

configure.ac
```diff
@@ -86,6 +86,31 @@ AS_IF([test "x$android" = "xno"], [
...
+AC_SUBST(GETTEXT_PACKAGE)
+AC_DEFINE_UNQUOTED([GETTEXT_PACKAGE], ["$GETTEXT_PACKAGE"],
+                   [The domain to use with gettext])
+
+AC_CONFIG_FILES([
+  Makefile
+  po/Makefile.in
```
12345ieee: Same for this.

12345ieee:
Indeed, you're right, they should not be installed.

You don't plan on localizing scanmem, I guess.

configure.ac
sriemer: Nope, that just sets a variable in `po/Makefile` and `po/Makefile.in` which is required for `intltool`. Can't get rid of `intltool` for non-GUI build unfortunately.

configure.ac
sriemer: `po/Makefile.in` is created from `po/Makefile.in.in` with `intltool`. Haven't found a way not to execute `intltool` for the non-GUI part. Will try again.

configure.ac
12345ieee: @sriemer , I tried it before commenting, and it worked. Inside the 'AS_IF'.

sriemer: @12345ieee Committing and a `git clean -fdx` helps here.
````
config.status: executing po/stamp-it commands
grep: po/Makefile.in: No such file or directory
config.status: error: po/Makefile.in.in was not created by intltoolize.
````

sriemer: It breaks the build without GUI.

12345ieee: Uhm, sorry it took a while, I indeed had leftovers from previous configures and removed manually some files, not a good approach.

I can get the gui-less build to go on if I also remove intltoolize from autogen.sh, but this will get complicated for no gain.

Let's remember that we call intltool even if not needed (either in a comment or the commit message) and let's call it a day.

sriemer:
Committed to `master` and `0.16.y`, thanks!

PR #251, opened 2017-07-17
Add a stop scan mechanism

12345ieee:
Adds a global variable, editable by the front-end and periodically read by the back-end during scan, to stop the current scan and retain only the values obtained till then.
Also adds the gui button to actually use the feature.

Obvious use case: scanning `firefox` takes a couple of minutes, if I got the value wrong I want to terminate the scan ASAP.

The checking is very coarse on purpose, to have minimal impact on the scan.

This is an important feature for me, planned for v0.17, as it changes the API.

@sriemer and @bkazemi , see any problem?

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -150,6 +150,7 @@ class GameConqueror():
         self.input_box = self.builder.get_object('Value_Input')
 
         self.scan_button = self.builder.get_object('Scan_Button')
+        self.stop_button = self.builder.get_object('Stop_Button')
         self.reset_button = self.builder.get_object('Reset_Button')
 
         ###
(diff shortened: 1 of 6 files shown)


bkazemi:
Looks good, but it'd be nice if we could get the feature into the CLI too (I actually use that, believe it or not ;-)). Maybe have a special case where a SIGINT causes the scan to stop rather than the entire program.

12345ieee:
@bkazemi , it's a good idea.

There is already some SIGINT handling in place for the `set` command (in `interrupt.h`), I imagine you can repurpose it to set this variable, so that we can keep a solution that works for every front-end.
When you have a working SIGINT handler feel free to open another PR.

I want merge this PR this evening, so I can go on to merge the rest of the gui stuff (which depends on this new button introduced here), unless somebody sees some issues.

bkazemi:
So I built this but it doesn't seem to be functioning. It looks like you missed a comma in `backend.py`'s BACKEND_FUNCS dict. Wonder why there was no Python backtrace...  Anyways, I fixed that and it still doesn't work. I am not familiar with GC, so this is as far as I go ;-)

12345ieee:
Thanks @bkazemi , indeed I forgot a colon.

It was fixed 3 commits later in my master by a related change.
This is why I'm in a hurry to merge all the backlog, so I can't overlook broken middle states.

The reason it might not work for you (it works here) it's because you have to point it to the newly compiled `libscanmem`. I use `LIBDIR = '/path/to/scanmem/.libs'` in `consts.py`, or you can run `make install`.

12345ieee:
I force-pushed, now it should work.

bkazemi:
Yep, that did it! I'll open a PR for the `scanmem` addition soon.

PR #252, opened 2017-07-20
libscanmem: fix grammatical errors in comments

bkazemi:
I've also changed single sentence comments to the seemingly more consistent lowercase style. American-style form has been used over British form (which I believe @sriemer said was preferred). I adjusted certain sentences slightly to make them clearer. These may need to be looked at to ensure I didn't alter a subtle meaning somewhere.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -42,7 +42,7 @@
 /*
  * sm_registercommand - add the command and a pointer to its handler to the commands list.
  *
- * so that free(data) works when destroying the list, I just concatenate the string
+ * So that free(data) works when destroying the list, I just concatenate the string
  * with the command structure. I could have used a static vector of commands, but this
  * way I can add aliases and macros at runtime (planned in future).
  *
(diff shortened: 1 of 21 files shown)


12345ieee:
Uhm, you change the first letter sometimes to uppercase and sometimes to lowercase, even in the same file (e.g. `targetmem.h`).

Either choose a consistent style or (maybe it's better) don't touch a line at all if initial case is the only change you make.

The rest looks ok, I'm not a native speaker, but the phrasing generally looks a lot less awkward.

bkazemi:
For multiple sentences I made sure it was capitalized. There was one exception with single sentences where if it starts with 'i', capitalize it. It seemed appropriate.

12345ieee:
Ah, now I see the logic behind it.
Seems ok for me, if @sriemer doesn't have any problems I'll merge it in a few days.

PR #253, opened 2017-07-20
Rest of the gui stuff

### LINKED ISSUE #248: Realtime memory editor
I wanted to know if it's possible to add sort of an auto-refresh (every sec for instance) to the memory editor. It would help identifying increasing/decreasing values (timers and such).

12345ieee:
Final PR for all the remaining gui stuff.

Good amount of cleanup, closes #248 and adds a full Italian translation.

I've tested this extensively, but extra pairs of eyes are always useful before I commit.


sriemer:
Thanks! Looks like nice improvements. I've only seen minor stuff like typos or white spaces at end of line when having a quick look what changed meanwhile.

12345ieee:
Thanks.

I know that auto-generated files (like the glade `.ui` file) tend to gain extra whitespace, but I prefer letting it be rather than having to scrub it manually at each commit.

PR #256, opened 2017-07-27
libscanmem assorted fixes/improvements: round 1

### LINKED ISSUE #208: `update` and non-numeric types
Steps:
- `option scan_data_type string` (or `bytearray`)
- `" some random string`
- get at least a match
- `update`

Expected:
Unless something changed in the process, everything remains the same.

What I get:

```
43> update 
.error: unsupported scan for current data type.
error: failed to scan target address space.
Please enter current value, or "help" for other commands.
0>
```

Numeric types work well.

### LINKED ISSUE #250: Detect_reverse_change & ineq_{forwards,reverse}
Sorry for the double email, missclick.

For who has seen my comments at cf89a4461d54e34a12044379e158b23a6245 , I was wondering about `detect_reverse_change`.

Currently, it's used together with `>` or `<` to mark also the values that did the opposite transition.
It has the following problems:

* the interaction with `!=` (and possibly other operators) is broken
* moving around the `ineq_{forwards,reverse}` flags is right in the middle of the hot scanning path
* this whole machinery basically duplicates the much simpler `!=`, if we exclude absurd edge cases like comparing float `NaN`s.

So, two proposals:

* remove the thing entirely, for a simpler and faster scan
* repurpose the mechanism to be used by `+N` and `-N`, where it can be of some utility, but if this is something we'd like to have, I think implementing a scan_function like the others would be better anyway

12345ieee:
This PR contains assorted fixes/improvements:

* Working `update`/MATCHANY for variable length types (closes #208 )
* Several speed ups and simplifications for busy scans (aka, 'nearly everything matches', sadly not the typical use case)
* Removal of the (IMHO) useless `detect_reverse_change`, as CHANGED has the same purpose (see commit for details), closes #250 .

As usual, see if you ( @sriemer , @bkazemi ) find some problems, otherwise I'll merge it in a few days.


value.h
```diff
@@ -164,16 +165,8 @@ static inline void truncval_to_flags(value_t *dst, match_flags flags)
 {
     assert(dst != NULL);
 
-    dst->flags.u64b &= flags.u64b;
```
bkazemi: Unfortunately with bit-fields the order is undefined; Therefore you can't know what value to store in `all_flags`. Also I don't think you had put anything in `all_flags` prior to the bitwise-or op so the flags were being set incorrectly (I think it would basically be optimized into a no-op, not sure). Numeric flags should probably be in a single variable..

12345ieee: `truncval()` should always be called (and it currently is, AFAICT), on correctly inited val and flags.

I know that bitfield order is implementation defined, but I was sure that all the bifields of a same type were equal.
In other words, `memset()/memcpy()` works, this has to work as well.

bkazemi: But isn't `truncval()` used for more than just zeroing the union?

12345ieee: Yes, the zero-er is another function.

This reimplementation might indeed do 6 "undefined" bit hands, if you didn't init the whole int16 for some reason, but the important point is that I should be guaranteed to get the right result on the flags, because they are at the same position in both match_flags.

Then I'll read from the single flags when I need them.

This is type punning, as I wrote, but it's the "right way (TM)" and it's fully supported, AFAIK.

I'm not a huge expert of these fringe behaviours, but I believe this one is fully portable (and, of course, it works).

bkazemi:
I keep forgetting this is a union 😥

This should work fine. I need sleep...

12345ieee:
Don't feel bad, I think I opened 2 different issues in the past because I had problems understanding that structure and I thought I found a bug.

You have all the time to take a look, no need to hurry, go get some sleep and come back refreshed tomorrow 😄.

bkazemi:
The naming is a bit strange in some places, surely. But that is another issue

12345ieee:
> Looks good to me; Only thing I was uncertain of is did you remove `detect_reverse_change` functionality? In the commit message you said it can be done with CHANGED, but I don't see how.

Should have said it better (I'll be more explicit): when you ask for a scan with `detect_reverse_change`, you can only use `DECREASED` or `INCREASED` and can only use it for subsequent scans.

What the scan does is use a pretty messy way to match a value if and only if `new > old || new < old` for INCREASED, or the other way around for DECREASED (which is actually the same anyway).

Now, discounting absurd edge cases (like float NaNs) `(new > old || new < old) == (new != old)`, so if you are looking for values that changed in any direction, you may as well use CHANGED (aka `!=`).

This simplifies flag handling by a lot (and speed in some cases), while basically sacrificing no functionality.

> Oh actually one more thing; There should be no need to distinguish between the number of graphemes and the number of bytes in a string, correct? Because if we do, then the member unification could stir up trouble edit: with multi-byte character encodings

To the scan code, a string scan isn't anything more that equality of the string codepoints byte by byte, so that works with no issue.
How do you get the codepoints you actually want in the uservalue string is another question, but as it is now, everything seems to work correctly, but I assume it's `readline` that guarantees that our `line` has the right bytes.
We use `strlen()` & friends, which AFAIK have no clue what a character encoding is, but the y count bytes, which is what they want.

E.g.
```
time echo -e 'option scan_data_type string\n" €€€€€€€€€€◕‿◕｡◕‿◕｡｡◕‿‿◕｡^̮^(◕‿◕)(｡◕‿◕｡)(｡◕‿‿◕｡)(^̮^)ʘ‿ʘಠ_ಠಠ⌣ಠಠ‿ಠ(ʘ‿ʘ)(ಠ_ಠ)(ಠ⌣ಠ)(ಠ‿ಠ)♥‿♥◔̯◔٩◔̯◔۶⊙﹏⊙ ̳ ̳ ̳ ̳ ͙ ڪ Ƹ̵̡Ӝ̵̨̄Ʒ(¬_¬)(；一_一)\nlist' | ./scanmem -p `pidof geany`
```
worked and gave me the 5 matches I expected (I pasted the same string in a new document)

bkazemi:
No wonder I had trouble wrapping my head around that code; From the description I assumed that it would check for `new > old` with the `<` command and vice-versa for the `>` command. Was I wrong with that assumption or was the implementation simply broken?

Yes, the only problem I could see with using byte length for strings is if we wanted to display the string size to the user, which it looks like we don't do. I think Unicode appears to work because your terminal (along with most Unix terminal these days) uses UTF-8 for Unicode support, which I assume you have. This is of course because UTF-8 supports using `char` type for characters which allows for old ASCII programs to work with UTF-8 assuming you only use ASCII range. Since @sriemer wanted to support UTF-16 as well, this will be a problem. But that's another problem, nothing to do with your commit.

12345ieee:
It checks for `new > old` with the `<` command *in addition* to the normal `new < old`, that's what makes it equal to `!=` (feel free to test it, past me was 99.9% sure of this).

The implementation worked mostly fine, but it did interact badly with other comparison operators (like `!=`), that would have needed to be `reverse_ineq` aware, and weren't.
All that was pretty convoluted and the extra flag handling was giving me a lot of headaches when interpreted as string/barray flag, so I decided to remove it altogether.

As for Unicode, we never, ever deal with character encoding in any part of the code, strings are printed as `char` sequences.
You're right, it works because my terminal is using the same encoding as my text editor (that should be UTF-8).

A fully unicode-aware libscanmem would have an option akin to the endian one, because you need to pass both the terminal/whatever input you use encoding and the target encoding.
I'm not really sure why you want to deal with all that, but I'd leave it for v0.18

bkazemi:
Yeah, the assumption I was making is that it would distinguish between the two. It looks like it did in `value.c`. It isn't possible to know if we have reversed inequality with `CHANGED`, so that was the difference.

That's true, but the problem is UTF-16 chars won't fit in a `char`. I personally don't need UTF-16, but @sriemer pointed out that some games use UTF-16 which would make scanmem choke.

12345ieee:
Yes, that's the difference, you know which direction the change went.
Honestly I have no clue what the use case for knowing that could be (the gui can't even get that info), but it's a very old feature, I imagine it predates CHANGED and INCRASEDBY and RANGE, there is really no use for it now.

As for UTF16, I get the point, while nothing stops the user from writing a simple UTF16toBArray converter and then looking for the BArray value sure it isn't great.

If you don't see any more issues, I think I'll merge it and start the next one, round 2 is cleanup time, I'll include your typo commit as well.

PR #257, opened 2017-07-30
libscanmem: implement multiple match deletions

bkazemi:
value.[ch]:
* Added a routine to parse unsigned integer sets.

targetmem.[ch]:
* Added a routine to get the last match id of the current matches.
  This is needed for the "open-ended" range `n..`.

handlers.[ch]:
* Implement support for the new set.
* Updated the `delete` documentation to reflect the changes introduced
  by this commit.

### DIFF
--- a/value.c
+++ b/value.c
@@ -240,6 +240,186 @@ bool parse_uservalue_float(const char *nptr, uservalue_t * val)
     return true;
 }
 
+#define DEFAULT_UINTLS_SZ (1000)
+
+/* XXX: this should be in a utility file of some sort (no pun intended) */
+static int _uint64_cmp(const void *i1, const void *i2)
+{
+    return (*(const uint64_t *)i1 - *(const uint64_t *)i2);
+}
+
+bool parse_uservalue_uintset(const char *lptr, uservalue_t *val)
+{
+    const char *tok, *tmpnum, *tmpnum_end;
+    bool        got_num;
+    uint64_t    last_num;
+    uint64_t   *valarr;
+    size_t      arr_szfilled, arr_maxsz;
+    size_t      vaidx = 0;
+    uservalue_t tokval;
+
+    assert(lptr);
+    assert(val);
+
+    enum {
+        NIL = -1,
+        NUM_TOK,
+        RNG_TOK,
+        CMA_TOK,
+    } last_type = NIL;
+
+    /* allocate space for value pointers */
+    if ((valarr = calloc(DEFAULT_UINTLS_SZ, sizeof(uint64_t))) == NULL)
+        return false;
+    if ((val->set_prop.n.one_to_val = malloc(sizeof(uint64_t))) == NULL ||
+        (val->set_prop.n.to_end_val = malloc(sizeof(uint64_t))) == NULL) {
+        free(valarr);
+        if (val->set_prop.n.one_to_val)
+            free(val->set_prop.n.one_to_val);
+        return false;
+    }
+
+    arr_szfilled = 0;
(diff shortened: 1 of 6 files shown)


bkazemi:
I got a touch impatient with the documentation updates, they probably need tweaking.

12345ieee:
I see you have you plugged the whole mechanism into `uservalue_t`, IMHO that struct is already too big for my taste for something that is involved in each scan_routine invocation, I would make an `uintset` struct with just what you need for this to work.

Also, I'd like to see all this in its file, because we are going to plug this into `dregion` as well, with some self-contained methods (like a way to walk the set that doesn't involve hardcoding what I want to do in the inner loop).

I imagine you had the idea of letting a user actually scan for an uintset in memory, which is why you used `uservalue_t`, I'm afraid that would be unbearably slow, but if it was your intention, I may run some tests.

Then, your `last_match_id`, can't you get the same info from `vars->num_matches`? If I understood what you're doing, they're guaranteed to return the same.

I'm in the middle of actually understanding it now, if you want to discuss in a more timely way, you find me at https://kiwiirc.com/client/irc.freenode.com/scanmem for a while.

value.c
```diff
@@ -240,6 +240,186 @@ bool parse_uservalue_float(const char *nptr, uservalue_t * val)
     return true;
 }
 
+#define DEFAULT_UINTLS_SZ (1000)
+
+/* XXX: this should be in a utility file of some sort (no pun intended) */
+static int _uint64_cmp(const void *i1, const void *i2)
+{
+    return (*(const uint64_t *)i1 - *(const uint64_t *)i2);
+}
```
12345ieee: I imagine this works because it's implicitly converted to `int` back at the function return.

Is this 100% guaranteed to work in all cases?

value.c
```diff
@@ -240,6 +240,186 @@ bool parse_uservalue_float(const char *nptr, uservalue_t * val)
...
+            goto error;                                                              \
+    } while (0)
+
+    for (tok = lptr; *tok; tok++) {
+        /* skip spaces */
+        while (tok && isspace(*tok))
+            tok++;
+
+        /* check if we just went over trailing space */
+        if (!tok || !*tok)
```
12345ieee: Is `!tok` even possible here?

value.c
```diff
@@ -240,6 +240,186 @@ bool parse_uservalue_float(const char *nptr, uservalue_t * val)
...
+
+    assert(lptr);
+    assert(val);
+
+    enum {
+        NIL = -1,
+        NUM_TOK,
+        RNG_TOK,
+        CMA_TOK,
+    } last_type = NIL;
```
12345ieee: Could you be a touch more descriptive? `NUM` is clear, but `RNG` clearly isn't a reference to randomness.

Ok, now I see what they are, the switch below told me. Maybe consider initing them as themselves, like `COMMA_TOK = ','` for clarity.

value.c
```diff
@@ -240,6 +240,186 @@ bool parse_uservalue_float(const char *nptr, uservalue_t * val)
...
+            continue;
+        case '.':
+            if (last_type == CMA_TOK || last_type == RNG_TOK)
+                goto error;
+            /* check for consecutive `..` */
+            if (tok != lptr && tok[-1] == '.')
+                last_type = RNG_TOK;
+            continue;
+        }
+
```
12345ieee: Stopping here for now.

12345ieee:
So, after a review and some discussion, me and @bkazemi have decided to go for a more flexible approach, with its own struct and file, which can be used for different handlers.

I'm closing this one, as the new one will be pretty different and it's better to start anew.

Issue #204, opened 2016-08-08
Deleted strings and bytearrays are recounted on rescan


12345ieee:
Steps:
- Attach to process
- Send `option scan_data_type string`
- Send `" <whatever>`
- Get N matches
- Run `delete 0`
- Send `" <whatever>` again

Expected:
Unless something happened in the process, N-1 matches, the ones I had after delete

What I get:
N matches, the one I deleted is back

Deleting multiple just has them all come back at rescan.

Numeric types behave correctly.

12345ieee:
Sorry, stray Return, issue body in next reply.

12345ieee:
Sorry for the extra noise, issue text for the email-bound:

------------------------------------------------------------------------------------

Steps:
* Attach to process
* Send `option scan_data_type string`
* Send `" <whatever>`
* Get N matches
* Run `delete 0`
* Send `" <whatever>` again

Expected:
Unless something happened, N-1 matches, the ones I had before delete

What I get:
N matches, the one I deleted is back

Deleting multiple just has them all come back at rescan.

12345ieee:
Got it, the problem is that the bytearray/string routine doesn't check the old value's flags. Even if the length is 0 (deleted match case), it still compares and reports, unlike numbers, which check both the new and the old flags.

We could either have the bytearray/string routine check that the data length is !=0, or check explicitly for deleted values before calling the testing routine.

Issue #186, opened 2016-06-29
libscanmem needs a clear separation from scanmem


sriemer:
The file `scanmem.h` contains all the common and global stuff but not all public interface definitions. This way no other e.g. C/C++ based front-end can be compiled.
`scanmem.h` should only contain the public interfaces of libscanmem so that a `libscanmem-dev` or `scanmem-devel` package can be created. The common and global stuff should be moved to a new header file `common.h`.

12345ieee:
Talking about the public API, I don't think `void reset_scan_progress()` should exist at all.

`scanmem` already resets the scan progress at the start of every scan and never checks the variable itself, so the function offers absolutely no functionality, apart screwing the scan counter, if called mid-scan.

sriemer:
Some parts done in `v0.16`. Some more work to do for `v0.17`.

sriemer:
`v0.17` should come with a completely cleaned up public API and `libscanmem2` then.

PR #259, opened 2017-08-08
Variable length types scan speedup

### LINKED ISSUE #226: Scan time of non-fixed-width types
Follows a very simple benchmark (against master):

```
andreas@trelitri:~/Progetti/scanmem$ time echo -e 'option scan_data_type int8\n0x64' | ./scanmem -p`pidof geany` 2>/dev/null
0> option scan_data_type int8
0> 100
136661> 
real    0m0,868s
user    0m0,788s
sys 0m0,076s

andreas@trelitri:~/Progetti/scanmem$ time echo -e 'option scan_data_type string\n" d' | ./scanmem -p`pidof geany` 2>/dev/null
0> option scan_data_type string
0> " d
136661> 
real    0m1,715s
user    0m1,648s
sys 0m0,060s

andreas@trelitri:~/Progetti/scanmem$ time echo -e 'option scan_data_type bytearray\n64' | ./scanmem -p`pidof geany` 2>/dev/null
0> option scan_data_type bytearray
0> 64
136661> 
real    0m1,752s
user    0m1,692s
sys 0m0,052s
```

in which we can see that looking for the **same exact 8 bits** with `int8` is twice as fast than doing it by non-fixed-width types. Same result for an int16 and a 2 char string.

Can we do something about this?

12345ieee:
I'm pulling in a rework of the variable length types (VLT for short) scan functions, salient points:

* Implement basic incoming memory length-checking, which offers a bit more safety and fixes the VLT returning from deletion of #204 .
* Massively optimize both small lengths (time -60%) and immediately not matching big lengths (-50%)'s scan time
* Rework bytearray wildcards, for another 15% speedup stacked on
* Finally closes #226 , as now VLT length is fully exploited

This PR features a fair share of type punning via pointer casting, I'm 99.9% sure that I've not introduced unaligned access on ARM, because the pointers I recast and dereference have enough alignment anyway, but I'd appreciate if someone could check on ARM.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -1260,7 +1260,7 @@ bool handler__watch(globals_t * vars, char **argv, unsigned argc)
         /* check if the new value is different */
         match_flags tmpflags;
         zero_match_flags(&tmpflags);
-        scan_routine_t valuecmp_routine = (sm_get_scanroutine(ANYNUMBER, MATCHCHANGED));
+        scan_routine_t valuecmp_routine = (sm_get_scanroutine(ANYNUMBER, MATCHCHANGED, NULL));
         if (valuecmp_routine(&o, &n, NULL, &tmpflags, address)) {
 
             valcpy(&o, &n);
--- a/handlers.c
+++ b/handlers.c
@@ -942,13 +942,14 @@ bool handler__default(globals_t * vars, char **argv, unsigned argc)
 {
     uservalue_t vals[2];
     uservalue_t *val = &vals[0];
-    bytearray_element_t *array = NULL;
     scan_match_type_t m = MATCHEQUALTO;
     char *ustr = argv[0];
     char *pos;
     bool ret = false;
 
     USEPARAMS();
+
+    zero_uservalue(val);
 
     switch(vars->options.scan_data_type)
     {
@@ -984,14 +985,7 @@ bool handler__default(globals_t * vars, char **argv, unsigned argc)
         break;
     case BYTEARRAY:
         /* attempt to parse command as a bytearray */
-        array = calloc(argc, sizeof(bytearray_element_t));
-    
-        if (array == NULL)
-        {
-            show_error("there's a memory allocation error.\n");
-            goto retl;
-        }
-        if (!parse_uservalue_bytearray(argv, argc, array, val)) {
+        if (!parse_uservalue_bytearray(argv, argc, val)) {
             show_error("unable to parse command `%s`\n", ustr);
             goto retl;
         }
@@ -1035,8 +1029,7 @@ bool handler__default(globals_t * vars, char **argv, unsigned argc)
     ret = true;
 
 retl:
-    if (array)
-        free(array);
+    free_uservalue(val);
 
     return ret;
 }
@@ -1572,18 +1565,11 @@ bool handler__write(globals_t * vars, char **argv, unsigned argc)
     case 1: // bytearray
         ; /* cheat gcc */
         /* call parse_uservalue_bytearray */
-        bytearray_element_t *array = calloc(data_width, sizeof(bytearray_element_t));
         uservalue_t val_buf;
-        if (array == NULL)
+        if(!parse_uservalue_bytearray(argv+3, argc-3, &val_buf))
         {
-            show_error("memory allocation failed.\n");
-            ret = false;
-            goto retl;
-        }
-        if(!parse_uservalue_bytearray(argv+3, argc-3, array, &val_buf)) 
-        {
-            show_error("bad byte array speicified.\n");
-            free(array);
+            show_error("bad byte array specified.\n");
+            free_uservalue(&val_buf);
             ret = false;
             goto retl;
         }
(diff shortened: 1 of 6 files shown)


12345ieee:
Did somebody have a look? 
I tried an ARM VM, but couldn't get any program to crash on unaligned access, so I can't tell if it would actully work.

sriemer:
Testing on physical HW would be best. A Raspberry Pi is ideal for that. The code looks good but I'd like to check on my Raspberry Pi 2.

12345ieee:
If you can, I'd really appreciate that, @sriemer .

The speed improvements should be evident in any architecture, assuming nothing crashes.

sriemer:
On the Raspberry Pi I see only small performance improvement. Nothing crashes but I also see a regression for strings. While a 7 char string is at 0.3s while scanning in a bash process, an 8 char string is at 1.7s. With the old code, that only happens from 9 chars onward. That needs to be fixed.

12345ieee:
Thanks for testing.

To be clear:

1-7: new is faster?
8: new is slower
9- : new is faster?

12345ieee:
Also, you're testing non-matching values, right? Otherwise all the speed gains are muddled in the much bigger time needed to add the match to the swaths.

12345ieee:
I just tested on my laptop non-matching strings of length 7,8,9.
In all 3 cases, sys time is the same and user time is roughly half with the new code.

Can you ensure that your kernel actually traps on misalignment, instead of executing with a speed penalty, via `/proc/cpu/alignment`?

sriemer:
Yes, of cause I'm testing non-matching stuff. Yep, also can't reproduce on x86_64.

Okay, I had "2 (fixup)" set before. I've set it to 4 and the cases which were slow now fail with SIGBUS. That means starting at 8 chars with this PR and starting at 9 chars with master.

Before:
```
User:           11239424
System:         0 (  (null))
Skipped:        0   
Half:           0   
Word:           0   
DWord:          11239424
Multi:          0   
User faults:    2 (fixup)
```
After:
```
User:           11239439
System:         0 (  (null))
Skipped:        0   
Half:           0   
Word:           0   
DWord:          11239424
Multi:          0   
User faults:    4 (signal)
```

12345ieee:
Ok, so it's clearly the cast to `uint64_t`, which happens in my PR starting at 8, but on master starting at 9.

It's clearly fault of the uservalue string, which is not aligned.

Can you test with bytearrays? I allocate them with `malloc()`, so they should be always aligned at 16 bytes boundary.

I'm also tagging this as "bug", as the current code fails on strings over 8.

sriemer:
The bytearrays work fine no matter which size.

12345ieee:
Great to know.

I'll prepare an extra fix to allocate strings the same way bytearrays are allocated.

Can you actually see the speed difference for bytearrays? Or even on amd64? Here the difference is big and I don't think it's an artifact of my setup.

sriemer:
Yes, the performance difference is big. I've now compiled my memfake tool with 200 MiB of zeroed memory to get a good measurement on my Raspberry Pi 2.

**bytearrays**
old: always 16.9s

new:
1..4 bytes: 10.7s            (37% less)
5..8 bytes: 12.5s
9..16 bytes: 14.8s
more bytes: time is gradually increasing until 16.9s where values are the same as with old scanmem

**strings**
old: 1..8 chars: 15.3s
new: 1..7 chars: 10.8s      (30% less)

12345ieee:
I've force-pushed said fix: 32637da

String scan speed got a ~5% faster on amd64 as well and now matches the bytearray time.

Please have a look and test it.

sriemer:
Hmm, it's a pity that Coverity doesn't find such stuff. With the new code everything is fine now. /proc/cpu/alignment doesn't complain about scanmem any more.

This can be merged now. Thanks!

sriemer:
I like the new "rebase and merge". 👍

12345ieee:
Yeah, it was about time github got a decent merge option from the site.

The only issue is that it adds you as committer and uses your default github email, which is not the same one I use for this project, so I keep merging manually.

12345ieee:
@sriemer , I've seen you didn't commit the string bugfix in the 0.16 branch.
Have you changed idea about maintaining 0.16 as a bugfix branch or did you simply forget?

sriemer:
Right, the 0.16 branch is the bugfix branch and should be maintained as that. I just didn't have enough time for it that day. Feel free to commit yourself.

12345ieee:
Pushed. I had to add 00cbbbb (pretty much purely cosmetic) as well, to get a clean cherry-pick.

sriemer:
Thanks, LGTM. I've committed 9534cdb8f89820e047c937fdde21b3aff41b6bec to show that this is not v0.16.1 any more but preparation for 0.16.2 instead.

12345ieee:
On the release topic, I'd like to accelerate a bit towards 0.17 as well.
Could you take a look at #262 ? #186 is the biggest roadblock towards 0.17 and I'd like some feedback on what we really want in `scanmem.h`, #262 should go in the right direction.

PR #261, opened 2017-08-16
Use pread64

ze00:
use pread64,plz!

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -434,7 +434,7 @@ ssize_t readregion(pid_t target, void *buf, size_t count, unsigned long offset)
     }
 
     /* try to honor the request */
-    len = pread(fd, buf, count, offset);
+    len = pread64(fd, buf, count, offset);
     
     /* clean up */
     close(fd);


bkazemi:
This will cause trouble with eventual BSD support as this is Linux-specific. glibc appears to deal with the function name change by treating the `pread` function as a functionally equivalent wrapper around `pread64`. Is there a specific reason you need this change?

ze00:
@bkazemi I use this lib on android,on android(32bit) pread isn't an alias for pread64.
I want read some data on offset 2634515104,if use pread,this will failed.

12345ieee:
@ze00 , I haven't looked around to find exactly where `pread` or `pread64` is supported, but this just looks like exchanging a problem for another problem.

If your platform needs special treatment (compared to linux), provide autoconf macros to check for what you need and then abstract away the difference in some reasonable way.

sriemer:
Yep, Android/bionic is special and we are missing contributors who regularly test latest scanmem code on Android. @12345ieee is right. Such stuff has to be determined at configure time. `configure.ac` already detects Android builds, does special handling for it, and with `AC_CHECK_FUNCS()` there is a macro which provides `HAVE_<FUNCTION>` defines after detecting if the function is available. That one could check if `pread64()` is available.

ze00:
@12345ieee @sriemer My test phone has 64bit cpu but runs 32bit android.I think problem stems from this.

12345ieee:
> My test phone has 64bit cpu but runs 32bit android.I think problem stems from this.

That's ok, both of these things can be detected at configure time.
If you provide the right `configure.ac` changes needed ( @sriemer pointed you to the needed macros above) we'll be happy to support your configuration, otherwise as it is now this is not merge-able.

Enyby:
> My test phone has 64bit cpu but runs 32bit android.I think problem stems from this.

https://github.com/scanmem/scanmem/blob/e77581aa328fd67a03f94ea12e875b383069d9e6/ptrace.c#L421

`unsigned long offset`, so offset is 32-bit and never contain value bigger from 4 GB.
Go on. pread accept type off_t which is signed. So offset is casted to it and treated as signed. So you can max use offset - 2GB.

Issue not in pread instead of pread64. Maybe if you replace it you avoid cast and get access to 4 GB. 

And connected thing - pread not always work with /mem file. Sometimes lseek and read - work, but pread  - not.

bkazemi:
@Enyby Great catch. Wouldn't of thought `off_t` was signed. Since the needed changes will make this PR irrelevant I'm going to close this. I'll be sure to credit you :-)

PR #262, opened 2017-08-24
Api simplification

12345ieee:
Some light restructuring to better separate libscanmem from scanmem (see #186 ).

I'm not sure what the best practices to present an interface are, but I'm trying to move away from `scanmem.h` as "it is included by everything and defines all the random things"

There's not much to review, just ensure I haven't borked the build in some edge case.


12345ieee:
Force-pushed, I had forgot an include guard.

PR #266, opened 2017-09-07
libscanmem: add sets.[ch], implement w/ delete

bkazemi:
```
sets.[ch]:
- New files add centralized user set parsing.

handlers.[ch]:
* Plug-in the new set parser.
* Updated the `delete` documentation to reflect the changes introduced
  by this commit.
```
So I've finished the new 'set' parser. @12345ieee if I misunderstood any of your requests, let me know.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -52,6 +52,7 @@
 #include "endianness.h"
 #include "handlers.h"
 #include "interrupt.h"
+#include "sets.h"
 #include "show_message.h"
 
 #define USEPARAMS() ((void) vars, (void) argv, (void) argc)     /* macro to hide gcc unused warnings */
@@ -450,44 +451,45 @@ out_free:
     return true;
 }
 
-/* XXX: handle multiple deletes, eg delete !1 2 3 4 5 6 */
 bool handler__delete(globals_t * vars, char **argv, unsigned argc)
 {
     size_t id;
-    char *end = NULL;
     match_location loc;
+    struct set del_set;
 
     if (argc != 2) {
         show_error("was expecting one argument, see `help delete`.\n");
         return false;
     }
 
-    /* parse argument */
-    id = strtoul(argv[1], &end, 0x00);
+    if (vars->num_matches == 0) {
+        show_error("nothing to delete.\n");
+        return false;
+    }
 
-    /* check that strtoul() worked */
-    if (argv[1][0] == '\0' || *end != '\0') {
-        show_error("sorry, couldnt parse `%s`, try `help delete`\n", argv[1]);
+    if (!parse_uintset(argv[1], &del_set, (size_t)vars->num_matches)) {
+        show_error("failed to parse the set, try `help delete`.\n");
         return false;
     }
-    
-    loc = nth_match(vars->matches, id);
-    
-    if (loc.swath)
-    {
-        /* it is not convenient to check whether anything else relies on this,
-           so just mark it as not a REAL match */
-        zero_match_flags(&loc.swath->data[loc.index].match_info);
-        vars->num_matches--;
-        return true;
+
+    foreach_set_bw(i, &del_set) {
+        id  = del_set.buf[i];
+        loc = nth_match(vars->matches, id);
+
+        if (loc.swath) {
+            /* It is not convenient to check whether anything else relies on this,
+               so just mark it as not a REAL match */
+            zero_match_flags(&loc.swath->data[loc.index].match_info);
+            vars->num_matches--;
+        } else {
+            show_error("BUG: delete: id <%zu> match failure\n", id);
+            set_cleanup(&del_set);
+            return false;
+        }
     }
-    else
-    {
-        /* I guess this is not a valid match-id */
-        show_warn("you specified a non-existant match `%u`.\n", id);
-        show_info("use \"list\" to list matches, or \"help\" for other commands.\n");
-        return false;
-    }
+    set_cleanup(&del_set);
+
+    return true;
 }
 
 bool handler__reset(globals_t * vars, char **argv, unsigned argc)
--- a/sets.c
+++ b/sets.c
@@ -1,0 +1,305 @@
+/*
+    Create sets.
+
+    Copyright (C) 2016-2017 Bijan Kazemi <bkazemi@users.sf.net>
+
+    This file is part of libscanmem.
+
+    This library is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License as published
+    by the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    along with this library.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include <sys/types.h>
+#include <ctype.h>
+#include <assert.h>
+#include <errno.h>
+
+#include "show_message.h"
+#include "sets.h"
+
+#define DEFAULT_UINTLS_SZ (64)
+
+/*
+ * -= The Set Data-Type =-
+ *
+ * A 'set' is a type similar in vein to the mathematical definition of a set.
+ * See: https://foldoc.org/set
+ *
+ * Sets support ranges. See the `delete` longdoc for more information.
+ *
+ * Restrictions of our set data-type:
+ * - No empty sets.
+ * - No duplicate elements.
+ * - All the elements must be of the same type.
+ */
+
+/*
+ * Used with `qsort()`.
+ *
+ * This function uses the less elegant branching solution
+ * to deal with possible overflow.
+ */
+static int _size_t_cmp(const void *_i1, const void *_i2)
+{
+    const size_t i1 = *(const size_t *)_i1, i2 = *(const size_t *)_i2;
+
+    if      (i1 <  i2) return -1;
+    else if (i1 == i2) return  0;
+    else               return  1;
+}
+
+/*
+ * Function used to grow the set array.
+ *
+ * We must use a separate pointer with realloc(), in case it fails. When that
+ * happens and we re-assign valarr, valarr is _not_ free()'d by realloc(), and
+ * valarr is then assigned to NULL, leaving the original valarr address lost.
+ */
+static inline bool inc_arr_sz(size_t **valarr, size_t *arr_maxsz, size_t maxsz,
+                              char **fail_reason)
+{
+    size_t *valarr_tmpptr;
+
+    if ( *arr_maxsz > maxsz / 2 ||
+        (valarr_tmpptr = realloc(valarr, (*arr_maxsz *= 2) * sizeof(size_t)))
+        == NULL)
+        if ((valarr_tmpptr = realloc(valarr, (*arr_maxsz = maxsz) *
+                                              sizeof(size_t))) == NULL) {
+            *fail_reason = "OOM (Out Of Memory)";
+            return false;
+        }
+
+    *valarr = valarr_tmpptr;
+
+    return true;
+}
+
+bool parse_uintset(const char *lptr, struct set *set, size_t maxsz)
+{
+    const char        *tok, *tmpnum = NULL, *tmpnum_end;
+    char              *tmpnum_endptr = NULL, *fail_reason = "BUG";
+    bool               got_num, is_hex = false;
+    size_t             last_num = 0, arr_szfilled, arr_maxsz;
+    size_t            *valarr;
+    size_t             vaidx = 0;
+    unsigned long long toknum;
+
+    assert(lptr && set);
+
+    /* ensure set is initialized */
+    memset(set, 0, sizeof(*set));
+
+    enum {
+        NIL = -1,
+        NUMBER_TOK,
+        RANGE_TOK,
+        COMMA_TOK,
+    } last_type = NIL;
+
+    /* allocate space for value pointers */
+    if ((valarr = malloc(DEFAULT_UINTLS_SZ * sizeof(size_t))) == NULL) {
+        show_error("%s(): OOM (Out Of Memory)\n", __func__);
+        return false;
+    }
+
+    arr_szfilled = 0;
+    arr_maxsz    = DEFAULT_UINTLS_SZ;
+    got_num      = false;
+
+    for (tok = lptr; *tok; tok++) {
+        /* skip spaces */
+        while (tok && isspace(*tok))
+            tok++;
+
+        /* check if we just went over trailing space */
+        if (!tok || !*tok)
+            break;
+
+        assert(arr_szfilled <= arr_maxsz); // should never fail
+        if (arr_szfilled == arr_maxsz && !inc_arr_sz(&valarr, &arr_maxsz,
+                                                     maxsz, &fail_reason))
+            goto error;
+
+        switch (*tok) {
+        case ',':
+            if (last_type == RANGE_TOK) {
+                fail_reason = "invalid range";
+                goto error;
+            }
+            last_type = COMMA_TOK;
+            continue;
+        case '.':
+            if (last_type == COMMA_TOK || last_type == RANGE_TOK) {
+                fail_reason = "invalid range";
+                goto error;
+            }
+            /* check for consecutive `..` */
+            if (tok[1] != '.') {
+                fail_reason = "bad token";
+                goto error;
+            }
+            last_type = RANGE_TOK;
+            tok++;
+            continue;
+        }
+
+        if (isdigit(*tok)) {
+            /* check for 0x hex prefix */
+            if (*tok == '0' && tolower(tok[1]) == 'x') {
+                if (!isxdigit(tok[2]))
+                    continue;
+                is_hex = true;
+                tok   += 2;
+            }
+
+            /* find the end of this number in string */
+            tmpnum_end = tok;
+            for ( ; isxdigit(*tmpnum_end); tmpnum_end++)
+                ;
+            tmpnum = strndup(tok, tmpnum_end - tok);
+
+            errno  = 0;
+            toknum = strtoull(tmpnum, &tmpnum_endptr, is_hex ? 16 : 10);
+            if (is_hex)
+                is_hex = false;
+            if (errno || *tmpnum_endptr != '\0') {
+                fail_reason = "strtoull() failed";
+                goto error;
+            }
+
+            free((void *)tmpnum);
+            tmpnum = NULL;
+
+            if (last_type == RANGE_TOK) {
+                if (!got_num) {
+                    /* we've got a {0 .. n} range */
+                    last_num  = toknum;
+                    last_type = NUMBER_TOK;
+                    got_num   = true;
+                    /* move token position to last number found (needed for 2+ digit nums */
+                    tok = tmpnum_end-1;
+                    /* bounds check */
+                    if (last_num >= maxsz) {
+                        fail_reason = "0..n range OOB (Out Of Bounds)";
+                        goto error;
+                    }
+                    /* pre-set the new size filled and grow array if necessary */
+                    if ((arr_szfilled += last_num+1) > arr_maxsz) // +1 for `0`
+                        while (arr_szfilled > arr_maxsz) {
+                            if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz,
+                                            &fail_reason))
+                                goto error;
+                        }
+                    /* fill up array with range */
+                    for (size_t i = 0; i <= last_num; i++)
+                        valarr[vaidx++] = i;
+                    continue;
+                }
+                if (toknum <= last_num || toknum >= maxsz) {
+                    fail_reason = "invalid range";
+                    goto error;
+                }
+
+                /* pre-set the new size filled and grow array if necessary */
+                if ((arr_szfilled += (toknum - last_num)) > arr_maxsz)
+                    while (arr_szfilled > arr_maxsz) {
+                        if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz, &fail_reason))
+                            goto error;
+                    }
+
+                /* fill up array with range */
+                for (size_t i = last_num+1; i <= toknum; i++)
+                    valarr[vaidx++] = i;
+            } else if (last_type == NUMBER_TOK) {
+                /* sanity check */
+                fail_reason = "impossible condition (last_type == NUMBER_TOK)";
+                goto error;
+            } else {
+                valarr[vaidx++] = toknum;
+                arr_szfilled++;
+            }
+            last_num  = toknum;
+            last_type = NUMBER_TOK;
+            if (!got_num)
+                got_num = true;
+            /* move token position to last number found (needed for 2+ digit nums) */
+            tok = tmpnum_end-1;
+        } else {
+            fail_reason = "bad token";
+            goto error;
+        }
+    }
+
+    /* check for {n .. end} range */
+    if (last_type == RANGE_TOK) {
+        if (!got_num) {
+            fail_reason = "invalid range";
+            goto error;
+        }
+        if (last_num >= maxsz) {
+            fail_reason = "n..end range OOB (Out Of Bounds)";
+            goto error;
+        }
+        if ((arr_szfilled += (maxsz-1 - last_num)) > arr_maxsz)
+            while (arr_szfilled > arr_maxsz) {
+                if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz, &fail_reason))
+                    goto error;
+            }
+        /* fill up array with range */
+        for (size_t i = last_num+1; i < maxsz; i++)
+            valarr[vaidx++] = i;
+    }
+
+    /* consider empty sets invalid */
+    if (arr_szfilled == 0) {
+        fail_reason = "empty set";
+        goto error;
+    }
+
+    /* sort the value array */
+    qsort(valarr, arr_szfilled, sizeof(valarr[0]), _size_t_cmp);
+
+    /* check if there are any duplicates */
+    for (size_t i = 0; i+1 < arr_szfilled; i++)
+        if (valarr[i] == valarr[i+1]) {
+            fail_reason = "duplicate element";
+            goto error;
+        }
+
+    /* check range (for individual entries, faster than checking each one) */
+    if (valarr[arr_szfilled-1] >= maxsz) {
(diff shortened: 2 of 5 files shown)


12345ieee:
Great timing, I'll review it right now.

You can find me here: https://kiwiirc.com/client/irc.freenode.com/scanmem as usual, if you need some real time interaction.

For posterity, this is a rewrite of #257 , after I asked for a rehaul of the concept.

sets.c
```diff
@@ -0,0 +1,222 @@
...
+
+#include "value.h"
+#include "sets.h"
+
+#define DEFAULT_UINTLS_SZ (1000)
+
+/*
+ * XXX: this should be in a utility file of some sort (no pun intended)
+ *
+ * This function uses the less elegant branching solution to deal with possible overflow.
```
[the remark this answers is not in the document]
bkazemi: It's more about what the 'guarantee' is. I figured, we want a set parser that accepts `size_t`'s regardless of the specifics of the user.

sets.c
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        if (isdigit(*tok)) {
+            /* find the end of this number in string */
+            tmpnum_end = tok;
+            for ( ; isdigit(*tmpnum_end); tmpnum_end++)
+                ;
+            tmpnum = strndup(tok, tmpnum_end - tok);
+
+            if (!parse_uservalue_int(tmpnum, &tokval))
```
12345ieee: It looks really weird here calling the uservalue function (and having to define that `get_val` over), when you'd only need a `strtoull()`. You can even drop the `value.h` include in that case, which would be appreciated.

sets.c
```diff
@@ -0,0 +1,222 @@
...
+                last_type = RNG_TOK;
+            continue;
+        }
+
+        if (isdigit(*tok)) {
+            /* find the end of this number in string */
+            tmpnum_end = tok;
+            for ( ; isdigit(*tmpnum_end); tmpnum_end++)
+                ;
+            tmpnum = strndup(tok, tmpnum_end - tok);
```
12345ieee: IMHO, this leaks `tmpnum`, as it lack a `free()`.

sets.c
bkazemi: Good idea. That's left over from when it was in `value.c` and it seemed appropriate then.

sets.h
```diff
@@ -0,0 +1,46 @@
...
+inline void set_free(struct set *set)
+{
+    if (set)
+        free(set->buf);
+    free(set);
+}
+
+/* iterate over set from front (forwards) */
+#define foreach_set_fw(i, set) \
+    for (size_t i = 0; i < (set)->sz; i++)
```
12345ieee: This is C11 only, you can't define vars in the `for` scope in C99.

Either make another subscope or, possibly for the better, drop the macros and leave looping to the caller, the interface is simple enough that hiding it does not really help much.

bkazemi: Vars in the for loop are a C99 extension. I think it is probably a good idea to keep the macros for the unsigned sets, as iterating backwards can be tricky.

12345ieee: Shame on me, I was 100% sure they were C11.
Ignore this whole comment block then.

sets.c
bkazemi: Fixed.

sets.c
```diff
@@ -0,0 +1,222 @@
...
+    assert(lptr);
+
+    enum {
+        NIL = -1,
+        NUM_TOK,
+        RNG_TOK,
+        CMA_TOK,
+    } last_type = NIL;
+
+#if __SIZEOF_SIZE_T__ == 8
```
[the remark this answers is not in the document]
bkazemi: Removed.

sets.c
```diff
@@ -0,0 +1,222 @@
...
+#include <ctype.h>
+#include <sys/types.h>
+
+#include "value.h"
+#include "sets.h"
+
+#define DEFAULT_UINTLS_SZ (1000)
+
+/*
+ * XXX: this should be in a utility file of some sort (no pun intended)
```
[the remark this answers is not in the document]
bkazemi: So you wanted this function in `sets.h`?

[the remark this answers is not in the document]
12345ieee: No, I'm saying that this comment made sense when the whole thing was in `value.{h,c}`.
Now that the set functions have their own file, this commit is unnecessary.

sets.c
```diff
@@ -0,0 +1,215 @@
...
+
+    /* consider empty sets invalid */
+    if (arr_szfilled == 0)
+        goto error;
+
+    /* sort the value array */
+    qsort(valarr, arr_szfilled, sizeof(valarr[0]), _size_t_cmp);
+
+    /* check if there are any duplicates */
+    for (size_t i = 0; i < (arr_szfilled % 2 ? arr_szfilled-1 : arr_szfilled); i++)
```
12345ieee: This guard is probably wrong, should be `i+1 < arr_szfilled`

sets.c
12345ieee: Yeah, looks good to me too.
I'll pass this whole thing under valgrind anyway when we're done tweaking.

sets.c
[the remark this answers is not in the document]
bkazemi: I was originally thinking it could be made a generic sort function for `qsort()` if we wanted to use that anywhere else in `libscanmem`. If you doubt we will use `qsort()` elsewhere, indeed I could leave it here.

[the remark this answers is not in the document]
12345ieee: Mah, leave it where it is, maybe just change the comment to say this function is used in `qsort()` only (though the signature alone is probably enough for anyone familiar with `qsort()`).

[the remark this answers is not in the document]
bkazemi: Updated

sets.c
bkazemi: It works, but this way is indeed cleaner (although an additional add op is introduced, it is negligible ;-)).  Will update

bkazemi:
Note to self: May be a good idea to put a description of a set in the C file.

sets.c
12345ieee: Thanks, a couple possible extra uops on a non-critical functions are worth it to avoid confusion every time I see this guard.

sets.c
```diff
@@ -0,0 +1,216 @@
...
+    else if (i1 == i2) return  0;
+    else               return  1;
+}
+
+bool parse_uintset(const char *lptr, struct set *set, size_t maxsz)
+{
+    const char        *tok, *tmpnum = NULL, *tmpnum_end;
+    char              *tmpnum_endptr = NULL;
+    bool               got_num;
+    size_t             last_num, arr_szfilled, arr_maxsz;
```
12345ieee: gcc 7 spits this:
```
sets.c: In function ‘parse_uintset’:
sets.c:54:24: warning: ‘last_num’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     size_t             last_num, arr_szfilled, arr_maxsz;
                        ^~~~~~~~
```
I know it's not actually used uninitialized, but please make it happy.

bkazemi: Right, I am paranoid I am missing some execution path but I think it's OK.

bkazemi: Fixed.

12345ieee:
```
48307311> delete 3.. 
error: failed to parse the set, try `help delete`

...

29> delete 3.. 
3> list 10
```

I get that you're OOM in the first case (I'm not sure why, 48M*8byte is just 400MB, I definitely have them), but please tell me that, not that I put in a bad set.

bkazemi:
Alright, I've added more helpful error messages. Try again and report the message. I can't seem to reproduce your second problem with ~300,000 elements, although now I get OOB when deleting over ~200,000 elements at once (32bit Linux 4G RAM)

12345ieee:
```
5239> delete 1200.. 
error: parse_uintset(): OOM (Out Of Memory)
error: failed to parse the set, try `help delete`
```

This looks a tad ridiculous, I definitely have 10kb of free RAM.

If you need, you find me on IRC as usual.

12345ieee:
It's not an OOM, it's most definitely L97.
If you don't need to duplicate the array you have to clamp to max size, not abort.

bkazemi:
Right, I noticed that, updated.

bkazemi:
Fixed hex parsing problems.

12345ieee:
Another bad failure:

```
5471> delete ,..5000 
error: parse_uintset(): BUG
```

bkazemi:
Fixed. That should be the last missed message. This error is intentional, I figured a `0..n` range should only be in the front of the set.

12345ieee:
Not good:

```
9023> delete .2
9022> exit

9023> delete .2.
9022> exit
```

Also pretty absurd (maybe it's a feature):

```
9023> delete ..2..
0> exit
```

handlers.c
```diff
@@ -450,44 +451,45 @@ bool handler__list(globals_t *vars, char **argv, unsigned argc)
...
-        show_warn("you specified a non-existant match `%u`.\n", id);
-        show_info("use \"list\" to list matches, or \"help\" for other commands.\n");
+
+    if (!parse_uintset(argv[1], &del_set, (size_t)vars->num_matches)) {
+        show_error("failed to parse the set, try `help delete`\n");
         return false;
     }
+
+    foreach_set_bw(i, &del_set) {
+        id  = ((unsigned *)del_set.buf)[i];
```
12345ieee: There is a nasty bug here.

```
14> list
[ 0] 559d0632e41a,  1 +          41a,  heap, 2, [I32 ]
[ 1] 7f22b0bb8300,  3 +         3300,  misc, 2, [I32 ]
[ 2] 7f22b0dde0a2,  6 +           a2,  misc, 2, [I32 ]
[ 3] 7f22b0dde120,  6 +          120,  misc, 2, [I32 ]
[ 4] 7f22b0dde45c,  6 +          45c,  misc, 2, [I32 ]
[ 5] 7f22b0dde570,  6 +          570,  misc, 2, [I32 ]
[ 6] 7f22b0dde5d1,  6 +          5d1,  misc, 2, [I32 ]
[ 7] 7f22b0ddeae0,  6 +          ae0,  misc, 2, [I32 ]
[ 8] 7ffd4cde3bb0,  7 +        1ebb0, stack, 2, [I32 ]
[ 9] 7ffd4cde3eb3,  7 +        1eeb3, stack, 2, [I32 ]
[10] 7ffd4cde3f7a,  7 +        1ef7a, stack, 2, [I32 ]
[11] 7ffd4cde4010,  7 +        1f010, stack, 2, [I32 ]
[12] 7ffd4cde4854,  7 +        1f854, stack, 2, [I32 ]
[13] 7ffd4cde4910,  7 +        1f910, stack, 2, [I32 ]
14> delete 1,2
12> list
[ 0] 7f22b0bb8300,  3 +         3300,  misc, 2, [I32 ]
[ 1] 7f22b0dde120,  6 +          120,  misc, 2, [I32 ]
[ 2] 7f22b0dde45c,  6 +          45c,  misc, 2, [I32 ]
[ 3] 7f22b0dde570,  6 +          570,  misc, 2, [I32 ]
[ 4] 7f22b0dde5d1,  6 +          5d1,  misc, 2, [I32 ]
[ 5] 7f22b0ddeae0,  6 +          ae0,  misc, 2, [I32 ]
[ 6] 7ffd4cde3bb0,  7 +        1ebb0, stack, 2, [I32 ]
[ 7] 7ffd4cde3eb3,  7 +        1eeb3, stack, 2, [I32 ]
[ 8] 7ffd4cde3f7a,  7 +        1ef7a, stack, 2, [I32 ]
[ 9] 7ffd4cde4010,  7 +        1f010, stack, 2, [I32 ]
[10] 7ffd4cde4854,  7 +        1f854, stack, 2, [I32 ]
[11] 7ffd4cde4910,  7 +        1f910, stack, 2, [I32 ]
12> exit
```

As you can see this actually deleted `0,2`, because you are reading the set, which contains `size_t` (int64) as `unsigned` (int32).
Endianness considered, this issued `delete 0; delete 1`, which then deleted 0 and 2 due to match shifting.

Please make the set pointer a `size_t`, as you fill it with `size_t`, there's no point in type erasure.
If we'll need a more general set, we'll see later.

If a `size_t` is used, the bug disappears.

12345ieee:
FYI, this is an example of followup patch to delete in O(N).
It's definitely not beautiful, but that's because there's no beautiful way to read the matches array.
It's not well tested, just to give an idea of what I'm aiming to.

``` diff
diff --git a/handlers.c b/handlers.c
index 87864f7..44b49bf 100644
--- a/handlers.c
+++ b/handlers.c
@@ -453,9 +453,7 @@ out_free:
 
 bool handler__delete(globals_t * vars, char **argv, unsigned argc)
 {
-    unsigned       id;
-    match_location loc;
-    struct set     del_set;
+    struct set del_set;
 
     if (argc != 2) {
         show_error("was expecting one argument, see `help delete`.\n");
@@ -472,24 +470,45 @@ bool handler__delete(globals_t * vars, char **argv, unsigned argc)
         return false;
     }
 
-    foreach_set_bw(i, &del_set) {
-        id  = ((unsigned *)del_set.buf)[i];
-        loc = nth_match(vars->matches, id);
+    size_t *limit_ptr = del_set.buf + del_set.sz - 1;
+    size_t *id_ptr = del_set.buf;
 
-        if (loc.swath) {
-            /* It is not convenient to check whether anything else relies on this,
-               so just mark it as not a REAL match */
-            zero_match_flags(&loc.swath->data[loc.index].match_info);
-            vars->num_matches--;
-        } else {
-            show_error("BUG: delete: id <%u> match failure\n", id);
-            set_cleanup(&del_set);
-            return false;
+    size_t match_counter = 0;
+
+    matches_and_old_values_swath *reading_swath_index = vars->matches->swaths;
+
+    unsigned int reading_iterator = 0;
+
+    while (reading_swath_index->first_byte_in_child) {
+        /* only actual matches are considered */
+        if (reading_swath_index->data[reading_iterator].match_info.all_flags != 0) {
+
+            if (match_counter++ == *id_ptr) {
+                /* It is not convenient to check whether anything else relies on this,
+                   so just mark it as not a REAL match */
+                zero_match_flags(&reading_swath_index->data[reading_iterator].match_info);
+                vars->num_matches--;
+
+                if (id_ptr++ == limit_ptr) {
+                    set_cleanup(&del_set);
+                    return true;
+                }
+            }
+        }
+
+        /* go on to the next one... */
+        ++reading_iterator;
+        if (reading_iterator >= reading_swath_index->number_of_bytes) {
+            reading_swath_index =
+                local_address_beyond_last_element(reading_swath_index);
+
+            reading_iterator = 0;
         }
     }
-    set_cleanup(&del_set);
 
-    return true;
+    show_error("BUG: delete: id <%u> match failure\n", *id_ptr);
+    set_cleanup(&del_set);
+    return false;
 }
 
 bool handler__reset(globals_t * vars, char **argv, unsigned argc)
diff --git a/sets.h b/sets.h
index 8488670..6355f16 100644
--- a/sets.h
+++ b/sets.h
@@ -25,8 +25,8 @@
 #include <stdlib.h>
 
 struct set {
-    size_t sz;  /* size of set (used) */
-    void  *buf; /* value buffer       */
+    size_t sz;   /* size of set (used) */
+    size_t *buf; /* value buffer       */
 };
 
 inline void set_cleanup(struct set *set)
```

handlers.c
bkazemi: Yeah I figured that cast was going to cause trouble eventually, however the problem is the `nth_match` matching (hehe) parameter is `unsigned` which of course isn't necessarily the same as `size_t`.

bkazemi:
Fixed the `.` bug and `..2..` is technically valid, so I guess it's a feature ;-)

bkazemi:
Another note, trimming the end of array off for `malloc()` might be good.

handlers.c
12345ieee: Relics from the 32bit time, when everything was simpler and longs were the same size as integers...
I also remembered you use a 32bit OS, so it's understandable you didn't catch it.

I'll add a treewide patch to change the matches indexing to `size_t` everywhere (since technically it could hold a whole program, hence the need for full object size).

Once I'm done, you can rebase and make the set.buf pointer a `size_t*`, this program has honestly too much type punning already, no need to introduce more.

handlers.c
```diff
@@ -450,44 +451,45 @@ bool handler__list(globals_t *vars, char **argv, unsigned argc)
...
         return false;
     }
 
-    /* parse argument */
-    id = strtoul(argv[1], &end, 0x00);
-
-    /* check that strtoul() worked */
-    if (argv[1][0] == '\0' || *end != '\0') {
-        show_error("sorry, couldnt parse `%s`, try `help delete`\n", argv[1]);
+    if (vars->num_matches == 0) {
```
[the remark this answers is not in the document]
12345ieee: In practice, yes, as the handlers are called by functions that put in the global address, which is never `NULL`.

In worst-case thinking, throw in an `assert()` and gain peace of mind.

handlers.c
12345ieee: As promised: b79a55b .

handlers.c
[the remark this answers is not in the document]
bkazemi: In this case I'll go ahead and leave it, as the opposite sounds like an impossible condition.

handlers.c
bkazemi: Excellent, pushed. 32bit will be relevant for a little longer, at least until our '07 laptops die completely ;-)

sets.c
```diff
@@ -0,0 +1,289 @@
...
+#include <string.h>
+#include <sys/types.h>
+#include <ctype.h>
+#include <assert.h>
+#include <errno.h>
+
+#include "show_message.h"
+#include "sets.h"
+
+#define DEFAULT_UINTLS_SZ (1000)
```
12345ieee: Allocating 1k `size_t` by default (4-8KB) seems a tad too much.
Something like 64 seems saner.

sets.c
```diff
@@ -0,0 +1,289 @@
...
+
+    enum {
+        NIL = -1,
+        NUMBER_TOK,
+        RANGE_TOK,
+        COMMA_TOK,
+    } last_type = NIL;
+
+    /* allocate space for value pointers */
+    if ((valarr = calloc(DEFAULT_UINTLS_SZ, sizeof(size_t))) == NULL) {
```
12345ieee: Do you need the memory zeroed? If not, use `malloc()`.
This is not for the laughable performance advantage, but to convey your intent.

sets.c
```diff
@@ -0,0 +1,289 @@
...
+    got_num      = false;
+
+/*
+ * NOTE: used locally. fragile, make sure to not use it with anything fancy.
+ *
+ * We must use a separate pointer with realloc(), in case it fails. When that
+ * happens, valarr is _not_ free()'d by realloc(), and valarr then points to
+ * NULL.
+ */
+#define INC_ARRAY_SZ                                                                           \
```
12345ieee: This thing keeps getting more and more complicated, it's probably time to make it a `static inline` function, so you can define local variables freely and make the structure clear.

sets.h
```diff
@@ -0,0 +1,51 @@
...
+    along with this library.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef SETS_H
+#define SETS_H
+
+#include <stdlib.h>
+
+struct set {
+    size_t  sz;  /* size of set (used) */
```
12345ieee: Can you please call the set member `size`?
If you prefer the abbreviate `sz` for the local variables of the set parser, that's fine, but I'd like a readable name in the rest of the program.

bkazemi: Changed.

sets.c
bkazemi: Changed.

sets.c
bkazemi: I suppose not. Changed.

sets.c
bkazemi: Changed.

handlers.c
```diff
@@ -450,44 +451,45 @@ bool handler__list(globals_t *vars, char **argv, unsigned argc)
...
+        id  = del_set.buf[i];
+        loc = nth_match(vars->matches, id);
+
+        if (loc.swath) {
+            /* It is not convenient to check whether anything else relies on this,
+               so just mark it as not a REAL match */
+            zero_match_flags(&loc.swath->data[loc.index].match_info);
+            vars->num_matches--;
+        } else {
+            show_error("BUG: delete: id <%z> match failure\n", id);
```
12345ieee: `%z` is a size specifier, not a full formatter. Here on Linux I need `%zu` for this to work.

12345ieee:
After the `%zu` change mentioned above (otherwise I see just `<%>`):

```
33> delete 0..4
error: BUG: delete: id <140463313849176> match failure
Please enter current value or "help" for other commands.
33> exit
```

sets.c
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+    /* check range (for individual entries, faster than checking each one) */
+    if (valarr[arr_szfilled-1] >= maxsz) {
+        fail_reason = "OOB (Out Of Bounds) element(s)";
+        goto error;
+    }
+
+    size_t *tmp_vaptr;
+    if ((tmp_vaptr = realloc(valarr, arr_szfilled)) == NULL) {
```
12345ieee: Here is missing a `* sizeof(size_t)`, which explains the absurd behaviour.

bkazemi: Good catch, pushed. Thanks

12345ieee:
This finally looks ready, thanks @bkazemi .

I'm adding on this minor patch, readability is subjective, but this will help future me for sure.

If you see nothing wrong, I'll merge.

``` diff
diff --git a/sets.c b/sets.c
index cc7cd71..b41b3b0 100644
--- a/sets.c
+++ b/sets.c
@@ -68,19 +68,19 @@ static int _size_t_cmp(const void *_i1, const void *_i2)
  * happens and we re-assign valarr, valarr is _not_ free()'d by realloc(), and
  * valarr is then assigned to NULL, leaving the original valarr address lost.
  */
-static inline bool inc_arr_sz(size_t **valarr, size_t *arr_maxsz, size_t maxsz,
-                              char **fail_reason)
+static inline bool inc_arr_sz(size_t **valarr, size_t *arr_maxsz, size_t maxsz)
 {
     size_t *valarr_tmpptr;
 
-    if ( *arr_maxsz > maxsz / 2 ||
-        (valarr_tmpptr = realloc(valarr, (*arr_maxsz *= 2) * sizeof(size_t)))
-        == NULL)
-        if ((valarr_tmpptr = realloc(valarr, (*arr_maxsz = maxsz) *
-                                              sizeof(size_t))) == NULL) {
-            *fail_reason = "OOM (Out Of Memory)";
-            return false;
-        }
+    if (*arr_maxsz > maxsz / 2)
+        *arr_maxsz = maxsz;
+    else
+        *arr_maxsz *= 2;
+
+    if ((valarr_tmpptr = realloc(valarr, (*arr_maxsz) * sizeof(size_t)))
+        == NULL) {
+        return false;
+    }
 
     *valarr = valarr_tmpptr;
 
@@ -129,9 +129,11 @@ bool parse_uintset(const char *lptr, struct set *set, size_t maxsz)
             break;
 
         assert(arr_szfilled <= arr_maxsz); // should never fail
-        if (arr_szfilled == arr_maxsz && !inc_arr_sz(&valarr, &arr_maxsz,
-                                                     maxsz, &fail_reason))
-            goto error;
+        if (arr_szfilled == arr_maxsz)
+            if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz)) {
+                fail_reason = "OOM (Out Of Memory)";
+                goto error;
+            }
 
         switch (*tok) {
         case ',':
@@ -199,9 +201,10 @@ bool parse_uintset(const char *lptr, struct set *set, size_t maxsz)
                     /* pre-set the new size filled and grow array if necessary */
                     if ((arr_szfilled += last_num+1) > arr_maxsz) // +1 for `0`
                         while (arr_szfilled > arr_maxsz) {
-                            if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz,
-                                            &fail_reason))
+                            if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz)) {
+                                fail_reason = "OOM (Out Of Memory)";
                                 goto error;
+                            }
                         }
                     /* fill up array with range */
                     for (size_t i = 0; i <= last_num; i++)
@@ -216,8 +219,10 @@ bool parse_uintset(const char *lptr, struct set *set, size_t maxsz)
                 /* pre-set the new size filled and grow array if necessary */
                 if ((arr_szfilled += (toknum - last_num)) > arr_maxsz)
                     while (arr_szfilled > arr_maxsz) {
-                        if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz, &fail_reason))
+                        if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz)) {
+                            fail_reason = "OOM (Out Of Memory)";
                             goto error;
+                        }
                     }
 
                 /* fill up array with range */
@@ -255,8 +260,10 @@ bool parse_uintset(const char *lptr, struct set *set, size_t maxsz)
         }
         if ((arr_szfilled += (maxsz-1 - last_num)) > arr_maxsz)
             while (arr_szfilled > arr_maxsz) {
-                if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz, &fail_reason))
+                if (!inc_arr_sz(&valarr, &arr_maxsz, maxsz)) {
+                    fail_reason = "OOM (Out Of Memory)";
                     goto error;
+                }
             }
         /* fill up array with range */
         for (size_t i = last_num+1; i < maxsz; i++)
diff --git a/sets.h b/sets.h
index 52c7dd1..7a558f2 100644
--- a/sets.h
+++ b/sets.h
@@ -42,11 +42,12 @@ inline void set_cleanup(struct set *set)
 /*
  * iterate over set from back (backwards)
  *
- * NOTE: `_zzaw2_df_` is reserved, named randomly to avoid namespace crash
+ * NOTE: `_reserved_for_set_iteration_zzaw2_df_` is reserved, named randomly to avoid namespace crash
  */
-#define foreach_set_bw(i, set)                                               \
-    for (size_t _zzaw2_df_ = 0, i = (set)->size-1; _zzaw2_df_ < (set)->size; \
-         _zzaw2_df_++, i--)
+#define foreach_set_bw(i, set)                                                \
+    for (size_t _reserved_for_set_iteration_zzaw2_df_ = 0, i = (set)->size-1; \
+         _reserved_for_set_iteration_zzaw2_df_ < (set)->size;                 \
+         _reserved_for_set_iteration_zzaw2_df_++, i--)
 
 bool parse_uintset(const char *, struct set *, size_t);
 #endif /* SETS_H */
```

bkazemi:
Looks good. However in
```
+    if ((valarr_tmpptr = realloc(valarr, (*arr_maxsz) * sizeof(size_t)))
+        == NULL) {
+        return false;
+    }
``` the parenthesis around `arr_maxsz` are unneeded.

12345ieee:
Merged as a3134f1 , with my last patch.

My 'delete in O(N)' patch will follow.

Now @bkazemi , for 0.18 , do you think it would be possible to support `!` as well?
Defined as `!N` == `..N-1,N+1..`, so that this can be plugged into `dregion` as well.

bkazemi:
Sure, that shouldn't be too hard to add.

PR #267, opened 2017-09-11
Incoming memory rewrite

12345ieee:
This PR completely rehauls the interface with which the process memory is processed by `searchregions()` and the scan functions.

The main commit of the bunch is 0e44526 , which has the rationale, a through list of changes and the advantages we obtain.

Some discussion points:
* A packed struct (`mem64_t`) is introduced to operate on 8-bytes chunks regardless of alignment, instead of copying them out each time. This *should* work (and passes a full `-fsanitize=address,undefined` run on amd64), but needs through testing on ARM (and maybe on i386).
* To provide a common `searchregion()` interface I had `ptrace()` fill the `data` buffer even in the `!HAVE_PROCMEM` case. If the extra consumed memory in this case is unacceptable, I'm open to alternative solutions, like refilling a smaller buffer multiple times.
* Likewise, the `peekbuf` cache now needs to be large enough to contain the largest allowed VLT. As of now, that's 64kB (aka `UINT16_MAX`), if it's deemed too much we can lower the maximum VLT width to something more reasonable.

The advantage of all this? As usual, it's scan speed, and also improved endianness correctness.

I've put @sriemer and @bkazemi as reviewers, as usual. I'm not committing this one until you guys are satisfied, please ask about any problem you see.

### DIFF
--- a/endianness.h
+++ b/endianness.h
@@ -31,7 +31,6 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include "scanmem.h"
 #include "value.h"
 
 /* true if host is big endian */
@@ -40,6 +39,11 @@ static const bool big_endian = true;
 #else
 static const bool big_endian = false;
 #endif
+
+static inline uint8_t swap_bytes8(uint8_t i)
+{
+    return i;
+}
 
 static inline uint16_t swap_bytes16(uint16_t i)
 {
@@ -86,9 +90,9 @@ static inline void swap_bytes_var(void *p, size_t num)
     return;
 }
 
-static inline void fix_endianness(globals_t *vars, value_t *data_value)
+static inline void fix_endianness(value_t *data_value, bool reverse_endianness)
 {
-    if (!vars->options.reverse_endianness) {
+    if (!reverse_endianness) {
         return;
     }
     if (data_value->flags.u64b) {
--- a/handlers.c
+++ b/handlers.c
@@ -278,7 +278,7 @@ bool handler__set(globals_t * vars, char **argv, unsigned argc)
                 while (reading_swath_index->first_byte_in_child) {
 
                     /* only actual matches are considered */
-                    if (flags_to_max_width_in_bytes(reading_swath_index->data[reading_iterator].match_info) > 0)
+                    if (reading_swath_index->data[reading_iterator].match_info.all_flags != 0)
                     {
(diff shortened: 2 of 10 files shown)


endianness.h
```diff
@@ -86,9 +90,9 @@ static inline void swap_bytes_var(void *p, size_t num)
     return;
 }
 
-static inline void fix_endianness(globals_t *vars, value_t *data_value)
+static inline void fix_endianness(value_t *data_value, bool reverse_endianness)
```
bkazemi: Woops, sorry!

That looks more natural to me.

12345ieee:
I went ahead and tested it on ARM32 myself, a 'raspberry pi zero w', with GCC 6.3.

The new code passed a through test with `-fsanitize=address,undefined` and delivered roughly the same relative speedups of the x86_64 version.

I checked the assembly myself, and the scanroutines indeed load the incoming memory one byte at a time, thanks to the `packed` magic.

I'll probably merge this in a few days, if you guys see no problems whatsoever.

bkazemi:
Sorry, I haven't much time to review. I'll have a thorough look tomorrow or the following day.

12345ieee:
That's ok, I'll wait.

bkazemi:
Really good work. I'll test this on x32 and get back, but functionally I don't see any problems.

bkazemi:
`set` is showing some issues.

I made a simple program that increments a variable every couple seconds.

The program starts with `int i = 0` on the stack; The address and value are printed.
```
0xbfe22678 1
0xbfe22678 2
0xbfe22678 3
```
```
1> list
[ 0] bfe22678,  3 +    1f678, stack, 2, [I32 I16 I8 ]
1> set 0=0
info: setting *0xbfe22678 to 0...
1> update
.ok
info: we currently have 1 matches.
1> list
[ 0] bfe22678,  3 +    1f678, stack, 3217300116, [I32 I16 I8 ]
````
```
0xbfe22678 -1077667179
```

**Comment by 12345ieee** [comment]
Great catch, @bkazemi .
The fix is trivial, I forgot a dereference in one of the tens of refactoring steps in `sm_setaddr()`:

``` diff
-        if (ptrace(PTRACE_POKEDATA, target, addr + i*sizeof(long), memarray+i) == -1L) {
+        if (ptrace(PTRACE_POKEDATA, target, addr + i*sizeof(long), memarray[i]) == -1L) {
```

Force pushed.

PR #268, opened 2017-09-21
handlers: check for matches properly

bkazemi:
Previously, we'd check if the `matches` struct is NULL, which may
be initialized when we have no matches.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -351,7 +351,7 @@ bool handler__list(globals_t *vars, char **argv, unsigned argc)
         }
     }
 
-    if (!vars->matches)
+    if (vars->num_matches == 0)
         return false;
 
     if ((v = malloc(buf_len)) == NULL)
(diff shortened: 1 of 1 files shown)


bkazemi:
Sometimes we need to check `matches` itself for initial scanning it seems.

12345ieee:
`checkmatches()` indeed dereferences the matches array without a check, while `searchregions()` does not.

Still, `num_matches > 0` implies `array != NULL`, so maybe just check for the num and add an assert for the array.

`handler__default()` has to be changed as well, as I said before.

bkazemi:
Added the check in `handler__default()`.

The problem is when `vars->matches != NULL` and `vars->num_matches == 0`. Because of this, we must check `vars->matches` itself as well for the initial search.

12345ieee:
Not really, `searchregions()` uses `allocate_array()`, which uses `realloc()`, which will happily accept any matches array (and even NULL) and destroy it.
So as of now, the checks are unnecessary, unless I'm mistaken.

If you still prefer the checks to be explicit, I'll live with it.

bkazemi:
It's more a style point rather than functional, for the `scanmem` UI.

12345ieee:
Ok, it's fine for me. I tested it and it obviously works.

Let's use github's shiny "Rebase and merge".

Issue #246, opened 2017-05-22
Option required to stay interactive after processing a script


sriemer:
I've received a scanmem feature request via e-mail that an option is required to stay interactive after processing commands in the scripting mode of scanmem. This way you could create your own aliases to set different defaults for scanmem.

Also the scripting mode should be documented in the man page as only few people know about it.

My proposal is introducing option `-i/--interactive` with description _"stay interactive after processing a script"_.

Looks like this:
`echo -e "$cmd1\n$cmd2" | scanmem -i`

What are your thoughts?

12345ieee:
While you're looking into this, I raise another point.
I use the `echo -e "$cmd1\n$cmd2" | scanmem`-thing to do fast debug/testing and it is pretty cumbersome, to say the least.
Also, a new `-i` flag when the interactive behaviour is the default seems pretty strange to me.

So I propose to rework the handling and turn the syntax into `scanmem -- $cmd1X$cmd2` where `X` is some reasonable separator, and then adding a `-q/--quit` flag, to have scanmem quit after processing the cmdline.

This is probably more work and also drops the pid grabbing without `-p`, but I wanted to know your opinion on it.

bkazemi:
Hmm, it would seem like a config file would be the best fit for persistent settings, maybe with something simple like libYAML. Should be easy to maintain, but it would add a new dependency.

I agree with @12345ieee that it seems a bit backwards to use a switch for interactive behavior as `scanmem` is described as a "interactive debugging utility" in the manpage. Unfortunately @12345ieee proposal would break backwards-compatibility if you were to disallow scripts through stdin, requiring a major version bump. But I also think it's a good idea for a one-line command separator, with `;` most likely for familiarity. But newlines should also be allowed as a separator for script files.

BTW - With modern shells like `bash` and `zsh` you can send a string to a command's stdin with a one-line heredoc:
```sh
scanmem <<<'cmd1\ncmd2'
```

sriemer:
What about a `-c` switch which can be repeated for providing commands which should be processed before getting interactive and which are ignored if there is input from stdin?

sriemer:
I'm sure @12345ieee can put the `exit` command at the end.

sriemer:
Or maybe it is best to use `-c` together with input from stdin to stay interactive after processing commands from stdin e.g. when reading commands from a file.

12345ieee:
> I'm sure @12345ieee can put the exit command at the end.

[...]

PR #269, opened 2017-09-23
main: Add initial commands parsing and execution

12345ieee:
Pretty easy solution for #246 .

I don't know if we need autotools magic to have `strtok_r()` available, or if you guys know of a suitable C-standard-only replacement.

### DIFF
--- a/main.c
+++ b/main.c
@@ -29,6 +29,7 @@
 #include "config.h"
 
 #include <unistd.h>
+#include <string.h>
 #include <stdio.h>
 #include <sys/types.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


bkazemi:
No worries, `strtok_r()` is a POSIX-compliant function.

12345ieee:
That I know, I'm just not sure if I need to add some horrific macro in `configure.ac` to check for it.

bkazemi:
All modern Unix OS's are POSIX-compliant (enough anyway), therefore there is no need to check.

main.c
```diff
@@ -176,6 +183,32 @@ int main(int argc, char **argv)
         show_user("Please enter current value, or \"help\" for other commands.\n");
     }
 
+    /* execute commands passed by `-c`, if any */
+    char *line;
+    char *saveptr = NULL;
```
bkazemi: It'd probably be cleaner to declare these inside the for loop.

bkazemi: It'd probably be cleaner to declare these inside the for loop.

PR #270, opened 2017-10-01
Fix a SEGV when checking `-c` commands

bkazemi:
This happens when you start `scanmem` without arguments.

### DIFF
--- a/main.c
+++ b/main.c
@@ -189,24 +189,26 @@ int main(int argc, char **argv)
     show_user_quick_help(vars->target);
 
     /* execute commands passed by `-c`, if any */
-    char *saveptr = NULL;
-    const char sep[] = ";\n";
-    for (char *line = strtok_r(initial_commands, sep, &saveptr);
-         line != NULL; line = strtok_r(NULL, sep, &saveptr))
-    {
-        if (vars->matches) {
-            show_user("%ld> %s\n", vars->num_matches, line);
-        } else {
-            show_user("> %s\n", line);
-        }
+    if (initial_commands) {
+        char *saveptr = NULL;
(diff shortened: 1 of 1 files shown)


bkazemi:
We need a new test suite ;-)

12345ieee:
What do you mean "without arguments"?
I tried `-c ' '`, `-c ';'`, `-c` and no c option at all and they all worked.

Let me guess, BSD's implementation of `strtok_r()` does not like its first argument to be `NULL`.

bkazemi:
No `-c` switch. This was on archlinux interestingly.

12345ieee:
Here everything works, regardless of other switches.
Anyway, merged with fa68bff , I edited the message a little to say it's an ArchLinux problem.

I definitely remember checking if `NULL` was fine, so I could avoid that `if`, really did not expect it to differ from Debian to AL.

As for a test suite, I'm working on it here: https://github.com/12345ieee/scanmem/tree/tests , but I was waiting for #267 to merge, or half the tests would be unbearably slow. Comments welcome.

bkazemi:
It was possibly caused by different versions of `glibc`, since I don't see any reference to this in the standard, I suppose it is considered implementation-defined.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/strtok.html

12345ieee:
I was actually wrong, with the glibc implementation I indeed dereferenced `NULL`:
https://github.com/bminor/glibc/blob/3d1870fa3301c5cd00d5fdab0014c4e22b71fef2/string/strtok_r.c#L46-L53
Oh, well, caught and fixed, I'll be more careful next time.

PR #272, opened 2017-10-02
Added spanish translation

salvadorp2001:
However, I was not able to test it because when trying to compile, it complains about
```config.status: error: cannot find input file: `po/Makefile.in.in'```

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -1,0 +1,530 @@
+# Spanish translations for Game Conqueror.
+# Copyright (C) 2009-2017 Wang Lu <coolwanglu(a)gmail.com>
+# This file is distributed under the same license as Game Conqueror.
+# Salvador Pardiñas <darkfm(at)vera.com.uy>, 2017.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: scanmem 0.17-pre\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2017-07-17 22:08+0200\n"
+"PO-Revision-Date: 2017-09-02 12:31+0300\n"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by 12345ieee** [comment]
Really? Have you run:
```
./autogen.sh && ./configure --enable-gui && make
```
with the needed dependencies installed?

File: po/es.po [hunk 1]
```diff
@@ -0,0 +1,530 @@
+# Spanish translations for Game Conqueror.
+# Copyright (C) 2009-2014 Wang Lu <coolwanglu(a)gmail.com>
+# This file is distributed under the same license as Game Conqueror.
+# Salvador Pardiñas <darkfm(at)vera.com.uy>, 2014.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: scanmem 0.14\n"
```
12345ieee: The metadata here are a bit messed up, current version is v0.16 and the revision date of this translation sure isn't in 2014.

salvadorp2001:
D'oh. I copied the metadata at the top from another translation and forgot to change dates.

12345ieee:
The file looks ok now.
I'm still a little puzzled by your error message, are you sure `autogen.sh` runs without errors?

salvadorp2001:
I'd forgotten to run autogen, then forgotten that Arch Linux doesn't install intltool by default. Then an error about not having PyGObject, then an error about "C Library doesn't support this locale". After fixing all that, yes, it seems to work and seems to be correct.

12345ieee:
Happy to hear it worked.

I'll merge it later, thanks for contributing!

12345ieee:
Merged as 328a7c6eab2c9c8dd .
Thanks for the translation.

Closing.

PR #275, opened 2017-10-03
Targetmem speedup

12345ieee:
I'm pulling in a series of reworks that dramatically speed up storing data in the matches array, for a total of roughly 55% (measure as snapshot time, for which this was the bottleneck).

* First commit is mostly cosmetic, but I'd like someone to check when I touch the `set` handler
* Second should fully rationalize includes, I'd like to be sure it compiles on other distros as well, speed neutral
* 3rd and 5th are improvements to `add_element()`
* 4th is the main part, a full rework of the `match_flags` type, same interface but much saner (IMHO) implementation, for correctness, portability, speed and no absurd union-derived problems (like having to track the active member, otherwise performance suffers).

Comments welcome.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -25,8 +25,6 @@
 # define _GNU_SOURCE
 #endif
 
-#include "config.h"
-
 #include <stdio.h>
 #include <stdlib.h>
 #include <getopt.h>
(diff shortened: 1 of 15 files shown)


bkazemi:
I agree 100% with using bitmasks. There is a good reason you see the majority of robust C projects using it over bitfields.

PR #280, opened 2017-10-05
Add new testing framework

12345ieee:
Add a very simple test script and configure the standard autotools facilities to run them.

The tests don't actually check the program output, but are intended to be used with gcc's sanitizers, to catch the sort of bugs that came up in #259 .

Add then the right configuration to `.travis.yml`, so that the tests run themselves automatically at each commit.

Adding for v0.17, comments and further testing ideas welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ missing
 py-compile
 stamp-h1
 tags
+test-driver
 
 # GameConqueror
 gui/GameConqueror.appdata.xml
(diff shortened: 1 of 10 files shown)


12345ieee:
Thanks @bkazemi , I wanted to add some runs under `valgrind`, to catch memory leaks/branch on uninitialized, but I don't have time to test all that before Friday.

I'll open an issue and target it to v0.18.

Merging this one as is.

PR #281, opened 2017-10-07
Add set data-type support to set command

bkazemi:


bkazemi:
Yeah, that rebase didn't work..

12345ieee:
Nah, it did work. Github doesn't really cope well with fancy rebases, but on the command line I can tell that it happened what I wanted.

I was waiting for you guys to review #279 , because I want it in before #275 , but it's simple enough that I'll probably merge it in by myself.

12345ieee:
A `set_cleanup()` is missing somewhere, this will leak a lot otherwise.

You can now rebase cleanly on scanmem/master, if you want.

bkazemi:
Instead of putting the `match_set` declaration at the top of the function and having to zero the struct twice (once in the initialization, and a `memset()` in `sets.c`) I just added it in the failure paths. In the future we should probably require the caller to zero the struct to avoid this.

12345ieee:
Eh, fair enough, for v0.17 it's enough.

I'm also not going to change it to O(N) right now, I risk introducing subtle bugs that we can't spot in less than a week.

I've adding a patch to `uservalue2value()` though, in the following commit, which ensures the whole value union is always set (otherwise the `show_info()` prints garbage) and does much less unnecessary work for values with many flags set.

There is an ambiguity issue with set (that was there before, as well), a value matched as `anynumber` and set to e.g. 'match-id=5' will have to decide between the double '5' or the integer '5'.
Currently (and in the prev master), the double wins, but in the future we might want to differentiate via `match-id=5` and `match-id=5.0`.

In as 61dc4ff and bf5309d , closing.

PR #283, opened 2017-10-10
Last code batch for v0.17

### LINKED ISSUE #187: [enhancement] Properly handle `A..B` when B<A
As of now, scanning for `A..B` when B<A will do all the work and yield exactly no results (as one would expect, the range [A,B] is empty, still, it's useless work).

Possible ways to improve on this behaviour:

* Tell the user that the range is empty beforehand, by putting a `A<B` check around here: https://github.com/scanmem/scanmem/blob/master/handlers.c#L935-L944
* Treat the range as [-inf, B) U (A, +inf], by inverting the matching checks (I couldn't figure out how this is done in the code)

### LINKED ISSUE #234: libscanmem: value.c: determining which uservalue_t flags to set
In https://github.com/scanmem/scanmem/blob/master/value.c#L209-L210 the int64_value and uint64_value members are unconditionally set which can cause undefined behavior/wrap-around when the numbers are out of range; thus, if a user supplied a number `x` where `x < INT32_MIN` there would be no reliable way to tell if the user indeed supplied a `int64_t` value or not, because both the signed and unsigned variables will be cast regardless. Before I opened a PR I wanted to confirm there was no subtle reason behind this.

12345ieee:
This and #169 should be the last code changes for 0.17, barring huge bugs.

cf593e0 : Closes #234 and sets the flags that a user would expect.

09b5792 63c399f 5cac3f2 : Close #187 , by introducing a reasonably robust checking mechanism.

d8a59b1 : QoL, I'm tired of not remembering if it's anynumber/number or double/float64 . Additional spellings welcome.


scanroutines.c
```diff
@@ -650,23 +650,53 @@ scan_routine_t sm_get_scanroutine(scan_data_type_t dt, scan_match_type_t mt, con
...
+    [FLOAT32]    = flag_f32b,
+    [FLOAT64]    = flag_f64b,
+    [BYTEARRAY]  = flags_max,
+    [STRING]     = flags_max
+};
 
 bool sm_choose_scanroutine(scan_data_type_t dt, scan_match_type_t mt, const uservalue_t* uval, bool reverse_endianness)
 {
-    const match_flags *uflags = uval ? &(uval->flags) : NULL;
+    match_flags uflags = uval ? uval->flags : flags_empty;
```
bkazemi: This could be `const`

handlers.c
```diff
@@ -903,17 +903,28 @@ bool handler__default(globals_t * vars, char **argv, unsigned argc)
...
+        if (pos) {
             *pos = '\0';
             if (!parse_uservalue_default(ustr, &vals[0]))
                 goto retl;
             ustr = pos + 2;
             if (!parse_uservalue_default(ustr, &vals[1]))
                 goto retl;
+
+            /* Check that the range is nonempty */
+            if (vals[0].float64_value > vals[1].float64_value) {
```
bkazemi: This is probably going to have the float problems we talked about.

scanroutines.c
12345ieee: I don't usually add `const` on non-pointer variables, otherwise more than half of the variables would need a `const`.

PR #287, opened 2017-11-14
handlers: $PAGER support in list command

bkazemi:


12345ieee:
A couple comments (even before I properly review it):

* You're probably doing too many things, split it in two commits
* A CLI user might or might not want a pager, but a library user never wants one. So before calling the pager in `list` one needs to check if we're in backend mode, like `show_user()` does.

bkazemi:
Because only `handlers.c` was edited, splitting the commit is a bit of a pain. I'll be sure to separate things in the future.

handlers.c
```diff
@@ -101,6 +101,9 @@ static FILE *get_pager(void)
     FILE *retfd = NULL;
     char *const emptyvec[1] = { NULL };
 
+    if (sm_globals.options.backend)
+        return stderr;
```
12345ieee: `list` needs `stdout` here.

You might need a `backend_pager` arg to `get_pager`.

handlers.c
```diff
@@ -432,11 +452,13 @@ bool handler__list(globals_t *vars, char **argv, unsigned argc)
...
     size_t reading_iterator = 0;
 
+    pager = get_pager();
+
     /* list all known matches */
     while (reading_swath_index->first_byte_in_child) {
         if (num == max_to_print) {
             if (num < vars->num_matches)
-                show_user("[...]\n");
+                fprintf(pager, "[...]\n");
```
12345ieee: Here the `show_user()` was used because I don't want to print the `[...]` in backend mode.

Check manually for `options.backend` here or something equivalent.

handlers.c
```diff
@@ -167,11 +170,27 @@ static FILE *get_pager(void)
...
+    signal(SIGPIPE, SIG_IGN);
+
     assert(retfd != NULL);
 
     return retfd;
 }
 
+static inline void close_pager(FILE *pager)
+{
+    if (pager != stderr) {
```
12345ieee: As said above, you need to add `stdout` here as well.

bkazemi:
Alright, I added a new feature to satisfy your request.

handlers.c
```diff
@@ -432,11 +456,18 @@ bool handler__list(globals_t *vars, char **argv, unsigned argc)
     matches_and_old_values_swath *reading_swath_index = vars->matches->swaths;
     size_t reading_iterator = 0;
 
+    if (ioctl(STDOUT_FILENO, TIOCGWINSZ, &w) == -1) {
+        show_warn("handler__list(): couldn't get terminal size.\n");
+        pager = get_pager(stdout);
+    } else {
+        pager = w.ws_row >= vars->num_matches ? stdout : get_pager(stdout);
```
12345ieee: Here you want `max_to_print`.
The warning fires in backend mode as well, but it doesn't really matter.

This whole block could also use an explanatory comment.

bkazemi: We want `vars->num_matches`, because `max_to_print` is likely always going to be bigger than the terminal line number. `max_to_print` also doesn't give the actual amount that would be outputted. I added a comment, though.

In backend mode we probably aren't attached to a terminal. I figured it was OK to leave it since we check for backend mode in `get_pager()`.

12345ieee:
Pushed with very minor changes as 0a2b7ed .

Closing, thanks for the patch.

PR #293, opened 2018-01-03
maps: fix spacing

bkazemi:
Multiple lines of code in sm_readmaps() were indented improperly.

This reduces a big diff with the BSD tree, because I caught this awhile ago working on that. Since it is unrelated to the BSD work, I opened this PR.
  
Edit: note that this is non-functional.


12345ieee:
Thanks, I had noticed it but forgot to fix it

Merged with 4925fbe54ef40c573918d5764ec63830a3c286b6

PR #294, opened 2018-01-08
cli: allow interrupts mid-scan

bkazemi:
Also:
* Add a new-line during an interrupt for cleaner output.
* s/intused/intr_used/ for clarity.
* New interrupt macro.

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -61,6 +61,7 @@
 #include "scanmem.h"
 #include "show_message.h"
 #include "targetmem.h"
+#include "interrupt.h"
 
 /* progress handling */
 #define NUM_DOTS (10)
@@ -305,6 +306,8 @@ bool sm_checkmatches(globals_t *vars,
     if (sm_attach(vars->target) == false)
         return false;
 
+    INTERRUPTABLESCAN();
+
     while (reading_swath.first_byte_in_child) {
         unsigned int match_length = 0;
         const mem64_t *memory_ptr;
@@ -367,7 +370,10 @@ bool sm_checkmatches(globals_t *vars,
                     print_a_dot();
                 }
                 /* stop scanning if asked to */
-                if (vars->stop_flag) break;
+                if (vars->stop_flag) {
+                    printf("\n");
+                    break;
+                }
             }
         }
         ++bytes_scanned;
@@ -383,7 +389,9 @@ bool sm_checkmatches(globals_t *vars,
             required_extra_bytes_to_record = 0; /* just in case */
         }
     }
-    
+
+    ENDINTERRUPTABLE();
+
     if (!(vars->matches = null_terminate(vars->matches, writing_swath_index)))
     {
         show_error("memory allocation error while reducing matches-array size\n");
@@ -458,6 +466,8 @@ bool sm_searchregions(globals_t *vars, scan_match_type_t match_type, const userv
         show_info("use the \"reset\" command to refresh regions.\n");
         return sm_detach(vars->target);
     }
+
+    INTERRUPTABLESCAN();
     
     total_size = sizeof(matches_and_old_values_array);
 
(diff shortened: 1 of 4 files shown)


12345ieee:
So, after some testing, everything works, great job!

From a code organization perspective, I'd prefer you made two different signal handler functions (what is now `void interruptable()`), one for each macro, so that you can "hardcode" if you want a jmp directly in the `INTERRUPTABLE*()` macroes.
This way you can easily reuse the macros anywhere, and you especially avoid including `sm.h` in `interrupt.h` to test on the globals.

I'm not an expert of signal handling and jumps, so if what I ask isn't technically feasible somehow, do tell me, we'll figure it out.

bkazemi:
Without the `scan_progress > 0` check, we can't be sure we are scanning or not. Plus we need `scanmem.h` anyways for the `sm_set_stop_flag()` function.

12345ieee:
This is what I had in mind: 

```diff
diff --git a/interrupt.h b/interrupt.h
index 0011d95..68fc82c 100644
--- a/interrupt.h
+++ b/interrupt.h
@@ -40,15 +40,18 @@ static unsigned intr_used;
 static void interrupted(int n)
 {
     (void) n;
-    printf("\n");
-    if (sm_globals.scan_progress > 0)
-        sm_set_stop_flag(true);
-    else
-        siglongjmp(jmpbuf, 1);
+    siglongjmp(jmpbuf, 1);
+}
+
+/* signal handler used to handle an interrupt during scans */
+static void interrupt_scan(int n)
+{
+    (void) n;
+    sm_set_stop_flag(true);
 }
 
 #define INTERRUPTABLE() ((oldsig = signal(SIGINT, interrupted)), intr_used = 1, sigsetjmp(jmpbuf, 1))
-#define INTERRUPTABLENOJMP() ((oldsig = signal(SIGINT, interrupted)), intr_used = 1)
+#define INTERRUPTABLESCAN() ((oldsig = signal(SIGINT, interrupt_scan)), intr_used = 1)
 #define ENDINTERRUPTABLE() (intr_used ? ((void) signal(SIGINT, oldsig), intr_used = 0) : (intr_used = 0))
```
with the macro called just before the scan loop start.
It looks a bit better to me, but sure not much.

Also, you shouldn't call `printf()` inside a signal handler, if you need that newline put it inside the `if (vars->stop_flag) break;`

bkazemi:
I see now. I changed it, but it doesn't seem to work for me. Let's see what comes up on the Travis build.

12345ieee:
I tested it, and it works for me.
But now with the double function we get annoying "unused function" bogus warnings.

It's time for `interrupt.c`, I guess.

Also, we probably need to declare `stop_flag` at least as `volatile`, if not even some weird atomic type.
Any ideas?

bkazemi:
The C file addition fixed the warning.

Volatile seems to be the way to go when modifying a global variable from an ISR. Added.

ptrace.c
```diff
@@ -463,6 +471,8 @@ bool sm_searchregions(globals_t *vars, scan_match_type_t match_type, const userv
         show_info("use the \"reset\" command to refresh regions.\n");
         return sm_detach(vars->target);
     }
+
+    INTERRUPTABLESCAN();
```
12345ieee: Shouldn't this go right before the `while (n) {`?
At the very least this block:
```
     vars->scan_progress = 0.0;
     vars->stop_flag = false;
     n = vars->regions->head;
```
has to go before it, or we risk leaving user-visible things in a mixed state.

bkazemi: Good point, not sure why I put it that far up.

bkazemi:
Signals are handled in the kernel's domain; They are always handled in some form whether you install a custom routine or not. No additional cost is added with a custom routine (except for whatever the interrupt routine function body does, of course).

12345ieee:
Ah, now you shouldn't need to include `sm.h` in `interrupt.h`, just in the `interrupt.c`.
Also, you can find me on chat for a bit.

Issue #288, opened 2017-11-15
Status of v0.17 in the distributions


12345ieee:
Let's have a look at which distros picked v0.17 up (table from #119 ) : 

| Distribution | Maintainer(s) | Latest Unstable | Link |
| --- | --- | --- | --- |
| Arch | Levente Polyak ( @anthraxx ) | 0.17-1 | https://www.archlinux.org/packages/community/x86_64/scanmem/ |
| Debian | Sebastian Parschauer ( @sriemer ),<br> **Uploader:** Kartik Mistry ( @kartikm ) | 0.17-2 | https://tracker.debian.org/pkg/scanmem, <br>devel: https://salsa.debian.org/debian/scanmem |
| Fedora | Igor Gnatenko ( @ignatenkobrain ), Jakub Hrozek  | 0.17-1 | https://apps.fedoraproject.org/packages/scanmem |
| Gentoo | **No maintainer** See: https://github.com/scanmem/scanmem/issues/288#issuecomment-360041311 | 0.16.1 | https://packages.gentoo.org/package/dev-util/scanmem |
| Mageia | Shlomi Fish ( @shlomif ) | 0.17-1 | https://madb.mageia.org/package/show/name/gameconqueror/release/cauldron |
| openSUSE | Sebastian Parschauer ( @sriemer ) | 0.17 | official: https://build.opensuse.org/package/show/openSUSE:Factory/scanmem, <br>with GUI: https://build.opensuse.org/package/show/home:sparschauer/scanmem |
| Ubuntu | see Debian | 0.17-2 | https://launchpad.net/ubuntu/+source/scanmem/ (depends on Debian version) |
| Ubuntu PPA | Sebastian Parschauer ( @sriemer ) | 0.17 | https://launchpad.net/%7Es-parschauer/+archive/ubuntu/game-cheating |

Only Gentoo is missing, see https://github.com/scanmem/scanmem/issues/288#issuecomment-360041311

ignatenkobrain:
yay, I'm the only one keeping scanmem up2date!

shlomif:
On Wed, 15 Nov 2017 20:37:14 +0000 (UTC)
Andrea Stacchiotti <notifications@github.com> wrote:

> So, a month after release, let's have a look at which distros picked v0.17 up
> (table from #119 ) : 
> 
> | Distribution | Maintainer(s) | Latest Unstable | Link |
> | --- | --- | --- | --- |
[...]

PR #296, opened 2018-01-20
Update the appdata to metainfo

### LINKED ISSUE #295: Lintian complains about legacy format in GameConqueror.appdata.xml
```
W: gameconqueror: appstream-metadata-in-legacy-location usr/share/appdata/GameConqueror.appdata.xml
E: gameconqueror: appstream-metadata-legacy-format usr/share/appdata/GameConqueror.appdata.xml
```

See: https://lintian.debian.org/tags/appstream-metadata-legacy-format.html

Blocks Debian packaging for #288.

sriemer:
Just fix #295.


sriemer:
@ignatenkobrain: I'm sure this one is interesting for you as well and this one is interesting for other distros in #288.

sriemer:
Okay, then my vote is for id `org.scanmem.gameconqueror` if we use non-existing domains. Is that fine for you?

12345ieee:
@sriemer , that's ok for me.

Considering that @kartikm uploaded to unstable earlier today (see https://tracker.debian.org/news/902673 ) I wonder if now we are committed to the initial name.

sriemer:
Nope, can do a 0.17-2.

12345ieee:
Good, let's go with `org.scanmem` as you said.
Update the PR and we can merge it.

12345ieee:
Many thanks for this PR.

I just noticed the policy is still called `org.freedesktop`, I guess we should change that as well, for consistency.

sriemer:
Yes, but that can be done for `v0.18`. I don't want to risk regressions for `v0.17.x` in this regard right now.

12345ieee:
Makes sense.

I'll open an issue later to remind us.

PR #297, opened 2018-01-21
Assorted cleanup

12345ieee:
Some random cleanup/cosmetic improvement I had lying around.

It has basically no functional impact, but comments welcome.

### DIFF
--- a/targetmem.c
+++ b/targetmem.c
@@ -123,13 +123,11 @@ match_location
 nth_match (matches_and_old_values_array *matches, size_t n)
 {
     size_t i = 0;
-
-    matches_and_old_values_swath *reading_swath_index = matches->swaths;
-
+    matches_and_old_values_swath *reading_swath_index;
     size_t reading_iterator = 0;
 
-    if (!matches)
-        return (match_location){NULL, 0};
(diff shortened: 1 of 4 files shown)


targetmem.c
```diff
@@ -122,15 +122,12 @@ void data_to_bytearray_text (char *buf, int buf_length,
 match_location
 nth_match (matches_and_old_values_array *matches, size_t n)
 {
-    size_t i = 0;
+    assert(matches);
```
bkazemi: Typically at the beginning of a block var declaration comes first, then function calls. I believe this is because C99 automatically hoists var decls to the beginning of a block, and ANSI C required it.

12345ieee: Yes, C99 allows you to define variables wherever you want.

Of course the stack space is still reserved at function entry, only the initialization is performed where the line is written.

I like using this feature, but the point of this commit was to move the NULL check **before** the array dereference (and transform it in an assert).

I can split declaration and initizliaztion and put all the declarations at the top if you guys like that better.

bkazemi: Yeah, that seems cleaner.

PR #299, opened 2018-01-22
scanmem: Add persistence of command history

12345ieee:
Add to `main.c` the ability to save and reload the readline history from a config file, in simple cases being `$HOME/.config/scanmem/history`.

I'm not sure this is ready to be merged, it works pretty well and I'd like to hear people's opinions.
Maybe some switches/env to control the file location, or some other behaviour?

### DIFF
--- a/main.c
+++ b/main.c
@@ -28,14 +28,25 @@
 
 #include "config.h"
 
-#include <unistd.h>
+#include <errno.h>
 #include <string.h>
 #include <stdio.h>
-#include <sys/types.h>
 #include <stdlib.h>
 #include <getopt.h>
 #include <stdbool.h>
 
+#include <pwd.h>
+#include <unistd.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+#if HAVE_LIBREADLINE
+# include <readline/history.h>
+#else
+# include "readline.h"
+#endif
+
+#include "common.h"
 #include "scanmem.h"
 #include "commands.h"
 #include "show_message.h"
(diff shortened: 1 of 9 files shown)


sriemer:
Let's do cross review. I review yours, you'll review mine.

Sounds like a good idea. But this should have a handy command to clear the history then and needs documentation in the man page.

12345ieee:
Hi, I'll have a look at yours shortly.
I agree with the missing documentation.

I'm a bit more perplexed on the ability to wipe the history, I limited the size to 1k lines, so it won't grow too big and you can always delete the file manually.

The issue with giving a `wipehistory` command, or even a `history {wipe|changefile|other_subcommand}` is that the CLI frontend needs to inject a command and its handler into libscanmem's commands list, as history is a frontend matter.

Considering that I want to add aliases as well, I could probably add a general purpose facility to "inject" commands with minimal extra work.

Opinions?

sriemer:
Looks mostly good. Thanks. Works as expected and clearing the history is not required for now.

I've checked the code. It has to be checked with Coverity Scan as it usually complains about `getenv()` being insecure. As we can use GNU extensions, we have to define `_GNU_SOURCE` before including `stdlib.h` and we can use `secure_getenv()` instead. Unfortunately, https://scan.coverity.com/ is down right now.

The commit description contains a superfluous space in "a lteration".

Correct K&R style is using: 
```
} else {
```
I also don't really like putting if statements and resulting actions into the same line. K&R style would be much better and consistent here. See: https://en.wikipedia.org/wiki/Indentation_style#K&R

K&R invented C and UNIX. So all GNU SW usually sticks to K&R.

Definition of `get_config_dir()` would be better with `get_config_dir(void)`.

sriemer:
Basically we should check if `secure_getenv()` is available first and avoid its use if it cannot be used e.g. for Android of BSD.

sriemer:
About `readline.c`: Non-void functions should always return something.

12345ieee:
I should have answered all your concerns @sriemer , beside:
* I directly repurposed `common.h` to be the common code carrier, I can rename it to `util.h` if you like it more.
* About the code style, if you dislike something feel free to attach a diff/correct it yourself before merging/pointing it out, I don't mind most style changes.
One of the few I care about is exactly K&R's single-line `} else {`, which I hate, but if you strongly prefer K&R I'll get a git hook to do the reformatting for me at commit time.

main.c
```diff
@@ -98,6 +109,57 @@ static inline void show_user_quick_help(pid_t target)
...
+
+    const char *config_dir = util_getenv("XDG_CONFIG_HOME");
+    if (config_dir == NULL) {
+        const char *home_dir = util_getenv("HOME");
+        if (home_dir == NULL) {
+            home_dir = getpwuid(getuid())->pw_dir;
+        }
+
+        size_t sm_config_dir_len = strlen(home_dir) + strlen("/.config/scanmem") + 1;
+        sm_config_dir = malloc(sm_config_dir_len * sizeof(char));
```
bkazemi: sizeof(char) is always 1, no multiplication needed.

main.c
```diff
@@ -98,6 +109,57 @@ static inline void show_user_quick_help(pid_t target)
...
+    else {
+        size_t sm_config_dir_len = strlen(config_dir) + strlen("/scanmem") + 1;
+        sm_config_dir = malloc(sm_config_dir_len * sizeof(char));
+        snprintf(sm_config_dir, sm_config_dir_len, "%s/scanmem", config_dir);
+    }
+
+    return sm_config_dir;
+}
+
+/* Recursive `mkdir()` implementation, from
```
bkazemi: Let's have this return a `bool` to be more consistent with libsm.

main.c
```diff
@@ -214,9 +276,23 @@ int main(int argc, char **argv)
...
+    if (vars->exit)
+        goto end;
+
+    /* Start of interactive mode: recover history from history file */
+    const unsigned int hist_max_size = 1000;
+    char *sm_config_dir = get_config_dir();
+    char *hist_file;
+
+    size_t hist_file_len = strlen(sm_config_dir) + strlen("/history") + 1;
+    hist_file = malloc(hist_file_len * sizeof(char));
```
bkazemi: NULL chk

PR #264, opened 2017-08-30
Full libscanmem separation

12345ieee:
Set of changes that should completely separate libscanmem from the CLI frontend, finally closing #186 .

This superseded #262 , which I closed earlier.

Highlights:

* Removed everything that's not a main `sm_*` function from `scanmem.h`
* Made both `menu.{h,c}` and `readline.{h,c}` sm dependences instead of libsm dependencies. libsm compiles without both. Complete independence of libsm from libreadline requires a bit more autotools meddling
* Expose only `sm_*` and `show_*` in the dynamic library, which seems a reasonable interface to maintain, instead of exposing basically everything
* Bring proper end-of-life cleanup to front-ends (the GUI used to hope that the OS would reclaim sm's memory at closure, now it actually calls `free()`)

The API still isn't final (that's why I didn't update the version number), as I have a last rather big change before v0.17, which changes a signature.

@sriemer , @bkazemi , please have a look, as I'm a bit outside my element here.

PR #302, opened 2018-02-16
scanmem: Introduce sub-command completion

sriemer:
Scanmem only completes the top-level commands with libreadline. But
some sub-commands like e.g. "scan_data_type" are really long and
often needed.
So make use of rl_line_buffer, rename the current completion function
to find_command(), move the index out of it, parse for spaces,
introduce a new data structure completion_t, use lists of lists,
and use recursion to complete sub-commands in arbitrary depth. Also
call find_command() within itself to provide "help" sub-command
completion.
For easy maintenance, define completion strings for relevant
commands, parse them for ',', '{' and '}' separators, and build up
the sub-command completion list in two levels from that.

This is the feature we've all been waiting for which stops typing "scan_data_type" with bleeding fingers all the time.


sriemer:
In v2 I've added function `sm_free_all_completions()` and I call it in `sm_cleanup()` to avoid memory leaks. I've also tested the `--without-readline` build and trying to break the completions with spaces at wrong locations like at the start or two of them in a row.

12345ieee:
So, I've tested it, and it's great! Thanks a lot for writing this.

I got the general gist of the nested completion search, I haven't examined the code in details, I'll have a closer look tomorrow when I'm fresher.

I tried to make it crash/leak in every way I could think of and I didn't succeed.

I ask you to please document in a big comment somewhere how to write a proper `HELP_COMPLETE`, I got that it has to be in this format `{a,...,{a...},...}` at most 2 levels deep, but future me is going to forget that.

While you're working on this, would it be possible to still work with multiple spaces?
`bash` autocompletes correctly regardless of starting spaces and amount of spaces between subcommands. 
scanmem already ignores initial and multiple spaces while tokenizing, it'd be great if the completer would work, too.

sriemer:
Thanks for your comment. Works with multiple spaces now. Was easy to implement with a `while` loop before calling a completion function. I've also added the big comment about the syntax. It is actually `"word1{word1.1,word1.2},word2{word2.1,word2.2}"`.

12345ieee:
This is looking great, I'll understand it in depth this evening.

Some points on the list functions visibility:
I introduced this "problem" with commit d619bc9 , part of #264 , in which I enforced full namespace separation of sm and libsm.
I had the same issue with the `show_*` functions, and I solved it the lazy way by exporting those symbols as well. I'm not a fan of my solution, but it was late in 0.17 and I didn't want to rehaul the file hierarchy further.

I think it's time to decide how to correctly handle common code between libsm and sm, I see 3 ways:

1) Compile everything in libsm and export all the common symbols - I'm not a fan of this, it'll cause namespace pollution of downstream project and it'll make maintaining a stable public API difficult
2) Move the logic of `show_message` in the header as well and include it where it's needed - as `show_message` is very small this looks like the simplest way, allows to drop its symbols from libsm
3) Build a dummy intermediate library of common units and link it statically into both sm and libsm - this affords a lot of freedom and it's probably the way to go if we were to share large chunks of code which need both a .h and a .c, but it looks overengineering for what we need now.

I'd go with 2 for now (moving `show_message` in another commit/PR, of course), opinions?

sriemer:
Yes, moving common stuff to the header is the way to go if functions are small any way. This project really started with inefficient programming with all functionality in .c files no matter how small or how often needed/called. E.g. the Linux kernel also just has a `list.h` with everything compiled inline.

Thanks for approval. I'll merge this.

PR #303, opened 2018-02-21
Introduce util_getenv() based on secure_getenv()

sriemer:
Check if `secure_getenv()` is available and define `util_getenv()` depending on that. On Android `secure_getenv()` is not available and `getenv()` is still used there instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,7 +11,7 @@ LT_INIT
 IT_PROG_INTLTOOL
 AM_PROG_CC_C_O
 
-AC_CHECK_FUNCS(memset strcasecmp strchr strdup strerror strtoul getline)
+AC_CHECK_FUNCS(getline secure_getenv)
 
 if test "x$ac_cv_func_getline" = "xno"; then
   AC_CHECK_FUNCS(fgetln)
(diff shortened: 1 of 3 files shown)


sriemer:
Basically this shouldn't be exported as `libscanmem` interface. Looks like we need new files `util.[ch]` where we put stuff which is shared between `scanmem` and `libscanmem` but not exported as `libscanmem` interface. So `util.h` should never be included from a header.

12345ieee:
Thanks for taking a look, didn't know about `secure_getenv()`.

About the mechanism used to share code, stuffing everything into an header works for this function, but doesn't work for `show_message`, because a couple functions access the `sm_globals`, which is defined in `scanmem.o` compilation unit.

For now you can leave this inside `scanmem.h` while #299 is cooking, but don't call it `sm_*` or it'll end up in the exported symbols, invent a prefix for the common code, maybe `util_*`.

This weekend I'll study a better way, building a dummy library, as I was saying in https://github.com/scanmem/scanmem/pull/302#issuecomment-366688523 , shouldn't too much hassle with autotools.

sriemer:
Okay, I've renamed the function to `util_getenv()`. Thanks. Anything left preventing merging for now?

12345ieee:
Considering that this is security-related it can be merged now.

When I'll change the code sharing mechanism I'll just move it in the right
place.

Issue #304, opened 2018-02-21
Performance penalty because of Meltdown/Spectre mitigation?


sriemer:
Latest kernels contain Meltdown/Spectre mitigations KPTI/KAISER and IBRS/Retpoline for the Intel CPUs. Always when we have to switch over to the kernel, then there is a performance penalty because of the kernel page table isolation.

Basics: https://imagefactory.otc.t-systems.com/Blog-Review/SpecExLeak/Benchmarks.html

When looking at the initial scan, then there is not much measurable performance impact. Basically there are only a couple of `pread()` calls which take most time. I've tested 1.5 GB of zeroed memory.

```
$ strace -c scanmem -p `pidof memfake` -c "10;exit"
...
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 99.59    0.614873      102479         6           pread
  0.40    0.002489         622         4           munmap
  0.00    0.000027           0        86           write
  0.00    0.000010           1        12           close
...
```

But what about further scans using `ptrace()` instead?

```
$ strace -c scanmem -p `pidof memfake` -c "0;10;exit"
...
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 99.47  117.200160           3  34855162         6 ptrace
  0.50    0.593829       98972         6           pread
  0.03    0.035552        2091        17           mremap
  0.00    0.000118          30         4           munmap
...
```

I had to interrupt this as it took too long. The high number of `ptrace()` calls just became even worse with latest Intel security bug mitigation variants. Without `strace`, the first test takes **10.9s** on my Haswell-EP CPU. It was around **10.7s** before. But the second test is extremely slow now. It takes **2min 42s** now because of so many `ptrace()` calls.

TODO: Test with kernel boot parameters `nopti` and `nospectre_v2`.

sriemer:
With `nopti nospectre_v2` the same test requires only **2min 5s**. I've repeated this twice and confirmed the **2min 42s** with the mitigations. So it takes 162s * 100 / 125s - 100% = **30% longer**.

So fixing #133 became even more important.

sriemer:
Tested kernel: 4.4.114-42-default on openSUSE 42.3.

12345ieee:
This is a really interesting benchmark.
[...]

Issue #133, opened 2015-10-09
ptrace: checkmatches() is very slow due to ptrace()


sriemer:
The issue is that checkmatches() only supports ptrace() but doesn't provide /proc/$pid/mem support in contrast to searchregions().

Let's compare some performance values taken from an Intel Core i5 system scanning in 500 MB of zeroed memory:

| Test | Search Pattern | Function | Time |
| --- | --- | --- | --- |
| nothing matches | 10 | searchregions() | about 10s |
| everything matches | 0 | searchregions() | about 20s |
| find nothing in matches | 10 after 0 | checkmatches() | about 60s |

Real world use case:
- search for bools

Bools are usually `unsigned char` 8-bit types with the only possible values 0 and 1. These values occur at many locations in memory. There are also large consecutive blocks of them - especially for 0.

sriemer:
TODO: Evaluate at which chunks size and which distance between matches which method is faster.

sriemer:
Postponed for now.

12345ieee:
On this topic: `sm_checkmatches()` wastes a lot of time out of the hot path by calling `truncval()` (inlined), which sets 10 flags one-by-one.

Commit 12345ieee@5f277fa introduces a value to set them all with a single int16 bitwise operation, for a 10% speedup. This could technically act on uninitialized padding bits if the receiving struct is uninitialized, but in any case the bits we're interested in get the right value.

12345ieee:
I have great news, what I was detailing in https://github.com/scanmem/scanmem/issues/304#issuecomment-367419399 works, and works very well.
Testing on 500 MB of zeroed memory, using `snapshot;update`, I recommend using this combination in memory retrieval tests, as it performs absolutely no matching logic.

`searchregions` alone
```
andreas@trelitri:~/Progetti/scanmem$ time ./scanmem -p `pidof memfake` -c 'snapshot;exit'
[...]
info: we currently have 524615680 matches.
524615680> exit

real	0m5,118s
user	0m4,166s
sys	0m0,908s
```

current master
```
andreas@trelitri:~/Progetti/scanmem$ time ./scanmem -p `pidof memfake` -c 'snapshot;update;exit'
[...]
info: we currently have 524615680 matches.
524615680> exit

real	0m57,418s
user	0m27,082s
sys	0m30,325s
```

this patch, with `READ_CHUNK` at 512 (1024 would shave another 1.5s, 2048 would shave 2.5s)
```
[...]

PR #305, opened 2018-02-25
Implement procmem support for `checkmatches`

12345ieee:
The middle commit description contains a performance analysis, that I report:

* Memory is read in chunks of 2048 bytes, using larger chunks
  does not help dense scans further
* Dense checkmatches time is dramatically decreased, around 3-4x,
  as the bottleneck was starting each syscall
* Sparse checkmatches time changed as follows:
  * For values not found in the cache, `pread` is ~30% slower than
    `ptrace`, this could be mitigated by smaller chucks, but it
    cannot be brought below 10% even with chunk size below 8
  * Therefore, relative timing depends on "cache occupancy" of
    each chunk (2048 bytes),
    at 1.3 values per chunk on average they break even

The cases in which this new patch is slower are really absurd, the effect becomes noticeable (> 1 sec) at around 1M values, so you'd need to scan 1M optimally spaced (> 2kB) values, for a total of 2 GB.

All the testing has been done on my PC, amd64 with Debian kernel 4.14.17-1.
I'm not sure if this kernel has all the spectre/meltdown mitigations, but the general picture isn't going to change.
About the chunk size, I left it at the beginning of the dense plateau, but we could consider raising it to the full 64kB and getting rid of some logic and the loop, as the `missing_bytes` cannot be more than 64kB and improve cache occupancy.

Speed in the `!HAVE_PROCMEM` case is clearly unchanged, the compiler has no issue doing the nested inlining.

This probably needs testing on arm, I'll see if I can manage in the following days.

Related: #133 , #304 .

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,7 +23,7 @@ if test "x$ac_cv_func_getline" = "xno"; then
 fi
 AM_CONDITIONAL([HAVE_GETLINE], [test "x$ac_cv_func_getline" != "xno"])
 
-AC_CHECK_HEADERS(fcntl.h limits.h stddef.h sys/time.h)
+AC_CHECK_HEADERS(fcntl.h limits.h stddef.h sys/ioctl.h sys/time.h)
 
 AC_FUNC_ALLOCA
 AC_FUNC_STRTOD
@@ -86,6 +86,14 @@ AS_IF([test "x$android" = "xno"], [
   # /proc/pid/mem is there but reading interesting data fails
   AC_DEFINE(HAVE_PROCMEM, [0], [Enable /proc/pid/mem support])
 ])
+
+# Allows disabling procmem support
+AC_ARG_ENABLE(procmem, [AS_HELP_STRING([--disable-procmem],
+                         [forcefully disable proc/pid/mem support])])
+AS_IF([test "x$enable_procmem" = "xno"], [
+  AC_DEFINE(HAVE_PROCMEM, [0], [Enable /proc/pid/mem support])
+])
+
 
 # Check for termcap and readline or bypass checking for the libraries.
 AC_ARG_WITH([readline], [AS_HELP_STRING([--without-readline],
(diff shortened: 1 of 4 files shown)


sriemer:
Thank you very much. The code looked good during review. But initial scan segfaults for me now at searching for `0` in 1500 * 1024 * 1024 bytes allocated by `calloc()`. Catching `SIGSEGV` in `sighander()` is a bad idea as it prevents the kernel message showing where it crashed and preventing core dump creation.

```
Core was generated by `scanmem -p 28384'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  allocate_enough_to_reach (swath_pointer_to_correct=<synthetic pointer>, last_byte_to_reach_plus_one=0x7fc3beaba018, array=0x0)
    at targetmem.h:137
137	    if (bytes_needed <= array->bytes_allocated) {
(gdb) bt
#0  allocate_enough_to_reach (swath_pointer_to_correct=<synthetic pointer>, last_byte_to_reach_plus_one=0x7fc3beaba018, array=0x0)
    at targetmem.h:137
#1  add_element (new_flags=flags_integer, new_byte=0 '\000', remote_address=0x7f7cb3a46002, swath=0x7fc2beabe02c, 
    array=0x603168 <sm_globals+8>) at targetmem.h:222
#2  sm_searchregions (vars=vars@entry=0x603160 <sm_globals>, match_type=<optimized out>, uservalue=uservalue@entry=0x7fff190a6e70)
    at ptrace.c:596
#3  0x00007fc51cdf422c in handler__default (vars=0x603160 <sm_globals>, argv=<optimized out>, argc=<optimized out>) at handlers.c:1087
#4  0x00007fc51cdefef7 in sm_execcommand (vars=0x603160 <sm_globals>, commandline=<optimized out>) at commands.c:283
#5  0x000000000040132c in main (argc=<optimized out>, argv=0x7fff190a71b8) at main.c:230
```

The `array` pointer is `NULL` and causes a NULL pointer dereference.

sriemer:
Never mind. I just ran out of memory on my laptop - happens also with the master branch. With 1GB it works as expected and the performance improvement is really nice.

sriemer:
Works as expected, code looks good, commits look good. So merging. Thanks!

PR #306, opened 2018-02-25
Add a utilities library

12345ieee:
As discussed in https://github.com/scanmem/scanmem/pull/302#issuecomment-366688523 , here is the dummy library implementation.

First commit just fixes the `HEADERS`, because new headers were added there by accident.

Second commit adds a library, `libutil.la`, that for now contains only `show_message.c`.
AM builds only the `.a` because of the `noinst` and then links it statically into both sm and libsm.

As a result the `show_*` functions can be dropped from the public API and everything still compiles.

Once this is merged I'll pick up #299 again, move `util_getenv` into its file and add said file to the `libutil_la_SOURCES`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,22 +18,22 @@ libscanmem_la_include_HEADERS = commands.h \
     maps.h \
     scanmem.h \
     scanroutines.h \
-    sets.h \
     show_message.h \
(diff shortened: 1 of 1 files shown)


sriemer:
LGTM, thanks!

PR #308, opened 2018-02-28
ptrace: Micro-optimize the comparison expressions in `sm_peekdata()`

12345ieee:
I'm putting this single commit up for review because I'd like to merge this speedup, but I wonder if this isn't actually a slowdown for somebody else with a different compiler.

I'm also not sure when it's worth making the expressions messier for a speedup, but this 10% looks significant enough if you guys see it as well.

----------------------------------

Replace the expressions used for the cache decision with
mathematically equivalent ones (as overflow is impossible in this case),
but that the compiler can optimize better.

On my system (gcc-7, amd64) this saves 10% time in a `snapshot;1` scan.

This optimization is really unstable, as equally simple and functionally
equivalent changes lead to slower code, so those ugly expressions
might be rewritten better once compilers get smarter.

Also sneak in a trivial fix for a71e1520, as `readmemory()`
returns 0 on failure, not -1.

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -234,7 +234,7 @@ extern inline bool sm_peekdata(const void *addr, uint16_t length, const mem64_t 
     /* check if we have a full cache hit */
     if (peekbuf.base != NULL &&
         reqaddr >= peekbuf.base &&
-        (unsigned long) (reqaddr + length - peekbuf.base) <= peekbuf.size)
+        peekbuf.base - reqaddr + peekbuf.size - length >= 0)
     {
         *result_ptr = (mem64_t*)&peekbuf.cache[reqaddr - peekbuf.base];
         *memlength = peekbuf.base - reqaddr + peekbuf.size;
@@ -242,7 +242,7 @@ extern inline bool sm_peekdata(const void *addr, uint16_t length, const mem64_t 
     }
     else if (peekbuf.base != NULL &&
              reqaddr >= peekbuf.base &&
-             (unsigned long) (reqaddr - peekbuf.base) < peekbuf.size)
+             peekbuf.base - reqaddr + peekbuf.size > 0)
     {
         assert(peekbuf.size != 0);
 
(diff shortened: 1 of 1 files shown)


ptrace.c
```diff
@@ -233,16 +233,16 @@ extern inline bool sm_peekdata(const void *addr, uint16_t length, const mem64_t
 
     /* check if we have a full cache hit */
     if (peekbuf.base != NULL &&
-        reqaddr >= peekbuf.base &&
-        (unsigned long) (reqaddr + length - peekbuf.base) <= peekbuf.size)
+        peekbuf.base - reqaddr <= 0 &&
```
bkazemi: Is this line really faster? Mine gives roughly the same speed with or without this change

12345ieee: The change is really really tiny, but it's consistently 1-2% faster this way.
The other line is the big one, ofc, that gives the remaining ~10%

Of course it depends on compiler and microarchitecture, your compiler may even output the same executable for both (mine doesn't).

Happy to hear it works for you as well, if the current line seems too obscure for you I can give up a 1% for clearer code.

bkazemi: Yeah, I think we should keep the original line for clarity. The rest looks good to merge

12345ieee:
Followed @bkazemi 's suggestion and merged.
Thanks.

12345ieee:
I tested it in ARM (32bit, gcc 6), and it does not work.
Crashes with a segfault at the beginning of `checkmatches()`.

I thought it was a sign conversion issue (on 64 bit `ptrdiff_t` wins over `unsigned int`, on 32 bit it's `unsigned int` that wins), but my simple fix didn't work, and I can't even explain how would it work for @bkazemi on x86.

It's late, so I'll have a better look tomorrow.
I'd rather not revert the commit, but if I can't fix it before the week end, I will.

bkazemi:
Huh. Which command caused the crash?

12345ieee:
A standard `snapshot;1` over 10 MB of memory.
Commit b0a9813e6f worked, so it's an issue I introduced in this commit.
I'm still convinced it's some silly signed/unsigned issue, but I find it very strange that it worked in x86 but didn't on ARM.

Time for some detailed investigation.

bkazemi:
I get the segfault now on x86. Must've been a problem on my end. My bad!

12345ieee:
I finally understood what caused the crash, see c899ac3 for a detailed explanation.
I ended up reverting the commit, as I couldn't find a correct implementation that also gained speed over the old one.

@bkazemi this commit only affected `checkmatches`, if you issued just the initial scan you couldn't have seen its effects.

PR #310, opened 2018-03-16
scanmem: centralize messaging

bkazemi:
Also:
* Remove an unnecessary include.

### DIFF
--- a/TODO
+++ b/TODO
@@ -2,8 +2,6 @@ Current
 =======
 * option: whether to search in readonly regions
 * reduce memory consumption (e.g. OVaMI padding)
-* completely centralize the printing in `show_message`, so that printing to
-  some other output stream is easy to add
 
 * see: https://github.com/scanmem/scanmem/issues for further TODOs
 
--- a/handlers.c
+++ b/handlers.c
@@ -30,13 +30,10 @@
 
 #include <stdio.h>
 #include <sys/types.h>
-#include <sys/wait.h>
 #include <sys/ioctl.h>
-#include <fcntl.h>
 #include <unistd.h>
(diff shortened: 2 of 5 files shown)


12345ieee:
Agh, that's not what I meant! Arguably I should have written it better.

I wanted to change every mention of `printf/fprintf/other_stuff that writes to console`, like https://github.com/scanmem/scanmem/blob/c899ac335c479680ab2980dc62b9a73db699dcde/handlers.c#L1176
to an appropriate `show_something()`.

Fow example I wanted some variant of `show_in_parser()`, so that the parser-getting code could be moved to `show_message`.

About the change made in this PR, I'm not a huge fan of it, same logic but more verbose.

bkazemi:
Fixed. I didn't change the `printf()` functions because data sent to `stdout` is different than "messages" sent to `stderr`.

12345ieee:
Sorry for disappearing for a while, between OS issues and Easter I was pretty busy.

I thought about it a bit, and as I said in chat, we really can't centralize calling the pager as much as I'd have liked, because you need full control on opening and closure.

So, I'll probably go for a much smaller patch that just moves the pager stuff in `show_message`.

I'm uncertain about `show_all()`, at this point it's probably clearer to show the full `fprintf(stderr,` given that we can't move everything to `show_*`.

I've pushed directly to your branch, @bkazemi , feel free to squash/modify as you see fit.

It should also be possible to remove some includes from `handlers.c` now, I leave that to you.

bkazemi:
Updated.

12345ieee:
Thanks, and sorry for having you rewrite it twice.

Merging.

PR #311, opened 2018-04-03
targetmem.h: replace uint with uint32_t

SSStormy:
This is causing compilation issues for projects using the scanmem.h API that do not have a uint typedef.


12345ieee:
Good catch, thanks.

The right(TM) type there is `unsigned int`: that index holds a number in [0, 8], the compiler can choose the integer it likes best, no need for us to specify it.

Closed by d77b171 .

PR #312, opened 2018-04-04
scanmem.h: add sm_reset API

justas-d:
This allows API users to avoid using `sm_init` and `sm_execcommand`just to use `sm_searchregions` and `sm_checkmatches` by allowing them to call `sm_reset` before `sm_readmaps`

I've also made sure that the "reset" command handler calls `sm_reset`

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -194,7 +194,7 @@ static inline void close_pager(FILE *pager)
     }
 }
 
-bool handler__set(globals_t * vars, char **argv, unsigned argc)
+static inline bool handler__set_unsafe(globals_t * vars, char **argv, unsigned argc)
 {
     unsigned block, seconds = 1;
     char *delay = NULL;
@@ -414,6 +414,14 @@ fail:
     ENDINTERRUPTABLE();
     return false;
     
+}
+
+bool handler__set(globals_t * vars, char **argv, unsigned argc)
+{
+    vars->scan_in_progress = true;
+    bool retval = handler__set_unsafe(vars, argv, argc);
+    vars->scan_in_progress = false;
+    return retval;
 }
 
 /* Accepts a numerical argument to print up to N matches, defaults to 10k
--- a/handlers.c
+++ b/handlers.c
@@ -617,17 +617,7 @@ bool handler__delete(globals_t * vars, char **argv, unsigned argc)
 bool handler__reset(globals_t * vars, char **argv, unsigned argc)
 {
     USEPARAMS();
-
-    /* reset scan progress */
-    vars->scan_progress = 0;
-
-    if (vars->matches) { free(vars->matches); vars->matches = NULL; vars->num_matches = 0; }
-
-    /* refresh list of regions */
-    l_destroy(vars->regions);
-
-    /* create a new linked list of regions */
-    if ((vars->regions = l_init()) == NULL) {
+    if(!sm_reset(vars)) {
         show_error("sorry, there was a problem allocating memory.\n");
         return false;
     }
--- a/ptrace.c
+++ b/ptrace.c
@@ -326,9 +326,9 @@ static inline uint16_t flags_to_memlength(scan_data_type_t scan_data_type, match
 
 /* This is the function that handles when you enter a value (or >, <, =) for the second or later time (i.e. when there's already a list of matches);
  * it reduces the list to those that still match. It returns false on failure to attach, detach, or reallocate memory, otherwise true. */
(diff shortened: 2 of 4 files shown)


12345ieee:
Honestly, avoid passing for `init` and `execcommand` at all might be a good idea for an external program.

The API was not written with that in mind (decoupling the library from the CLI is an ongoing process) so there is a lot of functionality tied to some commands, especially `option`, but it should be possible to change `sm_globals` directly and simulate all the option commands.

Then with `sm_reset`, `sm_readmaps` and manual tweaking of `sm_globals` you should indeed be able to scan. Then you'll have to iterate the matches array by replicating some of `targetmem`'s functionality.

About the code, public API should not have `globals_t *` as a parameter, `sm_globals` are meant to be a singleton with public API acting on it more or less directly, but I see that even the current API is inconsistent on that front.

So, I'd like to know a bit about your use case, so we can have an idea of what API a program that doesn't pass for `init`/`execcommand` need.

I'll open another issue to uniformize the public API, either all of them are going to have a `globals_t *` parameter or none of them.

SSStormy:
The use case here is python3 bindings for the `scanmem.h` API.  The inconsistent `globals_t *` usage was a pain which led me to mostly support APIs that take a `globals_t *`.  

The problem with that was that I could not find a valid way to use `sm_readmaps`, since it requires `list_t *regions` to be allocated and only `handler__reset` did that allocation.

From my point of view, I think that `globals_t *` should represent the "state" of an instance of scanmem, such that we could have multiple `globals_t *` attached to various PIDs. I have got a branch in the works that tries to do this in a way that doesn't break the old API but I doubt that will be possible due to `static struct peekbuf` in ptrace.c

SSStormy:
Just pushed a commit that adds `scan_in_progress` to `globals_t`. Here is the summary:

* `sm_reset` will return false and log using `show_info`.

* `handler__set` `sm_searchregions` and `sm_checkmatches` will set `scan_in_progress` to true before calling their respective`*_unsafe` version. After that procedure returns, `scan_in_progress` will be set to false and the return value of the `*_unsafe` method is returned. 

  This should combat bugs related with `scan_in_progress` being set when no scan is ongoing.

On the topic of `globals_t`: I'm on board with the idea of libsm being able to handle an arbitrary amount of globals. I'll try redesigning the API around this idea.

Anyhow, if there's anything else I can do to make this PR better, let me know.

12345ieee:
To talk about another issue, `sm_scan_routine` is another global variable, and keeping it global will not allow you to run 2 scans in parallel as well.

I think we cannot handle this in a bunch of unrelated PRs.
This is a substantial rework and we first have to get an idea of what API we want and what are the internal changes needed to get it (e.g. you'll need a `sm_list` for sure, probably some way to manipulate regions, e.g. `sm_dregions` and so on, basically most of the handlers could be extracted).

Your help on this is much appreciated, @SSStormy , I'll open a new issue shortly, so that we can discuss there about the new API and then go to implement the changes needed.

1rhino2:
Tested this properly. Original branch didn't build as-is, it predates the `get_pager`/`close_pager` move into `show_message.c/h`, so merging onto current main duplicated those functions. Not your bug, just staleness, removed the stale duplicates locally to get a clean build and confirmed the actual `sm_reset()` extraction + `handler__reset` still behave the same as before (region list refreshes correctly, `make check` and an ASan/UBSan rebuild both clean).

One design note on the `scan_in_progress` guard, since that's the whole point of this PR (letting an embedder call `sm_reset()` safely from outside the command loop): it's declared `volatile bool`, and `sm_reset()` does a plain check-then-set (`if (vars->scan_in_progress) return false; ... `) with no atomic compare-exchange or mutex. `volatile` only stops the compiler from caching/reordering the read, it doesn't make the check-and-act sequence atomic across threads. Two threads racing `sm_reset()` and `sm_searchregions()` at the same instant could both read `false` before either sets it to `true`, and the guard wouldn't catch it. If the target use case is genuinely multi-threaded (a GUI thread calling reset while a scan runs on a worker thread), it'd want `atomic_bool` + `atomic_compare_exchange_strong`, or a plain mutex around the whole check-and-flip. If it's meant for single-threaded embedders just guarding against reentrant calls from a signal handler or callback, the current approach is probably fine and I'd just double check that's the intended scope.

Everything else looks solid and the API itself (exposing `sm_reset()` for library consumers instead of only the CLI command) is a reasonable thing to want.

PR #314, opened 2018-04-15
ptrace: searchregions: Use a small buffer to store incoming memory

12345ieee:
The commit message (below) says it all.
This is a change to the memory handling, so I'd like to have some extra pairs of eyes on it.

-----------------------------

Instead of loading a whole region in memory, use a smaller
buffer (1 MiB) and multiple loads.
This massively improves memory consumed during the scan,
especially for large regions.

The searchregions loop has been rewritten and reoptimized,
and the dots printing with it.

Changes:
* Mild speedup (10%) for sparse scans, no change for dense
* The stop condition is checked every scanned MiB instead
  of every tenth of region, leading to more responsiveness
  where it matters
* Dots printing is checked every MiB, so when scanning very small
  regions (< 10 MiB) they will be bunched, but 10 MiB are so
  little that it isn't noticeable

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -543,47 +543,76 @@ bool sm_searchregions(globals_t *vars, scan_match_type_t match_type, const userv
 
     /* check every memory region */
     while (n) {
-        size_t nread = 0;
-        int dots_remaining = NUM_DOTS;
-        size_t bytes_at_next_dot;
+        size_t bytes_remaining;
         size_t bytes_per_dot;
         double progress_per_dot;
 
         /* load the next region */
         r = n->data;
         bytes_per_dot = r->size / NUM_DOTS;
-        bytes_at_next_dot = bytes_per_dot;
+        bytes_remaining = bytes_per_dot * NUM_DOTS;
         progress_per_dot = (double)bytes_per_dot / total_scan_bytes;
 
+/* The maximum logical size is a comfortable 1MiB (increasing it does not help).
+ * The actual allocation is that plus the rounded size of the maximum possible VLT.
+ * This is needed because the last byte might be scanned as max size VLT,
+ * thus need (2^16 - 2) extra bytes after it */
+#define MAX_BUFFER_SIZE (1<<20)
+#define MAX_ALLOC_SIZE  (MAX_BUFFER_SIZE + (1<<16))
+
         /* allocate data array */
-        if ((data = malloc(r->size * sizeof(char))) == NULL) {
+        size_t alloc_size = MIN(r->size, MAX_ALLOC_SIZE);
+        if ((data = malloc(alloc_size * sizeof(char))) == NULL) {
             show_error("sorry, there was a memory allocation error.\n");
             return false;
         }
 
         /* print a progress meter so user knows we haven't crashed */
-        /* cannot use show_info here because it'll append a '\n' */
         show_user("%02u/%02u searching %#10lx - %#10lx", ++regnum,
                 vars->regions->size, (unsigned long)r->start, (unsigned long)r->start + r->size);
         fflush(stderr);
 
-        /* read region into the `data` array */
-        nread = readmemory(data, r->start, r->size);
-        if (nread == 0) {
-            show_error("reading region %02u failed.\n", regnum);
-            free(data);
-            return false;
-        }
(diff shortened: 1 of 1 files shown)


12345ieee:
Can any of you guys ( @sriemer @bkazemi ) test this?

I'm pretty sure it works, but I don't trust myself enough to commit changes to memory management that have been tested only on my machine.

sriemer:
Thanks for the reminder. Okay, I'll look into this.

Btw.: I've also implemented caching with `ugtrain` by reading and writing 2K chunks. :smile: Most of the time dynamic memory objects can be read and written in a single chunk. Much nicer `strace` output there as well now. In contrast to `scanmem` I had to introduce many cache lists. ugtrain/ugtrain#52

ptrace.c
```diff
@@ -543,47 +543,76 @@ bool sm_searchregions(globals_t *vars, scan_match_type_t match_type, const userv
...
+                }
+
+                /* the whole region is finished */
+                if (memlength == 0) break;
+
+                /* stop scanning if asked to */
+                if (vars->stop_flag) break;
+
+                /* load the next buffer block */
+                size_t alloc_size = MIN(memlength, MAX_ALLOC_SIZE);
```
sriemer: Redefinition of `alloc_size` here which is misleading. Please rename this to `read_size`.

ptrace.c
```diff
@@ -543,47 +543,76 @@ bool sm_searchregions(globals_t *vars, scan_match_type_t match_type, const userv
...
         r = n->data;
         bytes_per_dot = r->size / NUM_DOTS;
-        bytes_at_next_dot = bytes_per_dot;
+        bytes_at_next_dot = bytes_per_dot * NUM_DOTS;
         progress_per_dot = (double)bytes_per_dot / total_scan_bytes;
 
+/* The maximum logical size is a comfortable 1MiB (increasing it does not help).
+ * The actual allocation is that plus the rounded size of the maximum possible VLT.
+ * This is needed because the last byte might be scanned as max size VLT,
+ * thus need 2^16 - 2 extra bytes after it */
```
sriemer: `2^16 - 2 bytes` sounds like a calculation and can be misleading. Please use `(2 extra bytes after it)`.

12345ieee: That IS a calculation. I'll write it as `(2^16 - 2)` for clarity.

ptrace.c
```diff
@@ -543,47 +543,76 @@ bool sm_searchregions(globals_t *vars, scan_match_type_t match_type, const userv
...
-        int dots_remaining = NUM_DOTS;
         size_t bytes_at_next_dot;
         size_t bytes_per_dot;
         double progress_per_dot;
 
         /* load the next region */
         r = n->data;
         bytes_per_dot = r->size / NUM_DOTS;
-        bytes_at_next_dot = bytes_per_dot;
+        bytes_at_next_dot = bytes_per_dot * NUM_DOTS;
```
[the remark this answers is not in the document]
12345ieee: Agreed.

ptrace.c
12345ieee: Agreed.

sriemer:
I've strace()ed this now. So you're reading in 1088 KiB (1 MiB + 64 KiB) chunks moving the offset 1024 KiB further. So 64 KiB (6.25%) are always read twice. And this doesn't have a performance impact?

sriemer:
I would like this better with 10 or 50 MiB chunks to reduce that ratio.

12345ieee:
Thanks for having an in-depth look.

Indeed a `memcpy()` approach would be a ton slower, having `pread()` do overlapping reads is the only
efficient way.

As I have written in a comment `The maximum logical size is a comfortable 1MiB (increasing it does not help)`, I have tried measuring performance improvements by changing the buffer size, but IIRC already at 512kB I saw no differences against bigger buffers.

If you have a system that benefits from the increased buffer size, or technical reasons why a ratio under 1% is preferred, I'll be happy to rise it to the best size that offers performance benefits, otherwise I don't see why use more memory for no apparent benefit.

sriemer:
Okay, leave it as 1 MiB. Not even on a Raspberry Pi 2 with slow memory I can measure a clear difference.

PR #318, opened 2018-06-13
Fix version number to 0.18~dev

sriemer:
We are in scanmem/GameConqueror 0.18 development already. So the
version numbers should reflect that in configure and the man pages.
0.18~dev is less than 0.18 for distribution packages due to special
character '~'.


sriemer:
Obviously, the linker is wrong in the Travis CI build.

12345ieee:
I should have done this just after the 0.17 tag, how did I forget?
Thank you for fixing it.

I've been a little absent lately, busy with other stuff.

I've the seen the build failure, it's clearly on travis' side, as I retriggered an older one that succeeded and now it fails.

I'll have a look at travis in a couple days, feel free to merge this anytime.

Issue #319, opened 2018-06-14
Travis CI build fails


sriemer:
There is a gcc-6 bug in the Ubuntu toolchain PPA:
https://stackoverflow.com/questions/50024731/ld-unrecognized-option-push-state-no-as-needed

12345ieee:
So exactly our setup (recent gcc with UB sanitizer) is broken, what a joy.

I've seen you try some variations under the `travis` branch, did you have any success?

sriemer:
Nope, no success so far. A Bionic build worker should fix this but Bionic is not available. No idea how to change the linker with autotools.

PR #324, opened 2018-10-13
.travis.yml: Use docker and Ubuntu Bionic for gcc-7

sriemer:
The gcc option "-fsanitize=address,undefined" requires at least gcc-6
but Travis-CI only provides Ubuntu Trusty environments. The hack to
use a testing toolchain for gcc-6 and binutils 2.26 does not work
any more.
So use docker with the ubuntu:bionic image for the regular build.
It provides gcc-7 as the default gcc.

Reference: #319

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,46 @@
 language: c
 sudo: required
-compiler: gcc-6
+
+services:
+  - docker
 
 env:
   global:
   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
   #   via the "travis encrypt" command using the project repo's public key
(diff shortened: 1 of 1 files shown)


sriemer:
Hmm, I should get rid of the double git clone. It is too visible in the traffic graph. There should be a possibility to share a directory between host and container.

sriemer:
The issue is fixed now.

V2 changes:
* move all docker stuff to `script:`
* pass the git directory from the host to the container and set it as working directory
* remove double git clone and surrounding commands verifying that it worked

sriemer:
V3 changes:
* export ASAN_OPTIONS='halt_on_error=1'; export UBSAN_OPTIONS='halt_on_error=1'; make check ...
* reduce duplication with environment variables for Docker commands

Setting environment variables in the Docker container is not trivial. Here I've set them in the same shell as `make check` as they are used by the sanitizers at run-time. No further changes planned. Looks good to me now. Sorry for submitting too quickly.

12345ieee:
Have you tested a coverity scan push as well?

sriemer:
Not yet, Coverity Scan has issues in general right now - at least with manual submission. Will push to the coverity_scan branch now.

sriemer:
The coverity_scan build works and Coverity Scan analyzed it. Coverity passed. :+1:

sriemer:
Merged, thanks! Will rebase the other PR.

PR #346, opened 2019-08-08
Added command for reading value from memory

mjdr:
read < type > < addr >

### DIFF
--- a/handlers.h
+++ b/handlers.h
@@ -330,6 +330,21 @@ bool handler__dump(globals_t *vars, char **argv, unsigned argc);
 
 bool handler__write(globals_t *vars, char **argv, unsigned argc);
 
+#define READ_COMPLETE VALUE_TYPES
+#define READ_SHRTDOC "read value of a specific memory location"
+#define READ_LONGDOC "usage: write <value_type> <address>\n" \
+                "\n" \
+                "Read <value_type> from <address>\n" \
+                "<value_type> should be one of:\n" \
+                "\tint{8|16|32|64} (or i{8|16|32|64} for short)\n" \
+                "\tfloat{32|64} (or f{32|64} for short)\n" \
+                "\n" \
+                "Example:\n" \
+                "\tread i16 60103e\n" \
+                "\tread float32 60103e\n" 
+
+bool handler__read(globals_t *vars, char **argv, unsigned argc);
+
 #define OPTION_COMPLETE "scan_data_type{number,int,float," VALUE_TYPES \
     "},region_scan_level{1,2,3},dump_with_ascii{0,1},endianness{0,1,2}"
 #define OPTION_SHRTDOC "set runtime options of scanmem, see `help option`"
--- a/handlers.c
+++ b/handlers.c
@@ -1615,6 +1615,109 @@ retl:
     return ret;
 }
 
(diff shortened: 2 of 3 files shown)


sriemer:
If you know address and type already, for what do you need `scanmem` then?
There is the `dump` command. Also `ugtrain` can watch a configured variable in memory before starting to modify it.

handlers.h
```diff
@@ -330,6 +330,21 @@ bool handler__dump(globals_t *vars, char **argv, unsigned argc);
 
 bool handler__write(globals_t *vars, char **argv, unsigned argc);
 
+#define READ_COMPLETE VALUE_TYPES
+#define READ_SHRTDOC "read value of a specific memory location"
+#define READ_LONGDOC "usage: write <value_type> <address>\n" \
```
sriemer: `write` is a copy and paste mistake here.

12345ieee:
I'm not convinced this should be separate from the `dump <addr> <len>`, but if we want it just to give a nice presentation of the read data we should call refactor it and `dump` so that they share most of the code, not repeating it in both places.

sriemer:
How would you handle ASLR/PIC/PIE or dynamic memory with this?

superprogram:
hello all, i am beginer on linux

i have run scanmem,
but im stuck now here.

Enter the pid of the process to search using the "pid" command.

PID PROCESS, example i have browser Chromium and there inside like to use SCANMEM!

any can help?

12345ieee:
@superprogram This is not the right place for such requests, this is for code discussion.
If you are a newbie, use gameconqueror, that has a user friendly interface.

1rhino2:
Found a real bug here by actually running it, not just reading. `handler__read`'s `bool ret;` is never assigned on the success path, only in the error branches (`ret = false; goto retl;`). On a successful read it falls straight through the `if (st == ...)` print chain into `retl:` and returns whatever garbage was on the stack for `ret`.

That's not just a style nit, it breaks scripted usage. Repro: `scanmem -p <pid> -e -c "read i32 <addr>;lregions"`. The `read` succeeds and prints the value, but `sm_execcommand()` gets back the uninitialized `ret`, which reads falsy often enough that with `-e` (exit-on-error) the rest of the chain silently never runs, `lregions` just doesn't happen, no error message either. Same command without `-e` "works" because the failure just triggers the quick-help hint instead of aborting, which is how I traced it back. Fix is just adding `ret = true;` before falling into `retl:` on the success path.

Smaller stuff while I was in there:
- `get_pager(stdout)` is called unconditionally at the top before argument validation, so a malformed `read` command (bad type, bad address, wrong arg count) still forks a pager subprocess for nothing.
- output has no trailing newline, runs into the next prompt line.
- `parse_scan_data_type(argv[1])` gets called twice, the second call right before the address parse is redundant.
- `datatype` is computed but always 0 and unused (`(void) datatype;`), looks like leftover copy-paste from `handler__write`.
- indentation mixes tabs and spaces in a few spots, worth a pass with whatever the rest of the file uses (4-space).

Feature itself is useful, `write` has an inverse `read` missing today, worth having. Happy to take a pass at the `ret` fix and cleanup if that's useful, it's a small patch.

PR #353, opened 2019-10-08
treewide: Fix printf format issues

### LINKED ISSUE #351: Coverity Scan reports new defects (printf format issues)
It is basically rather minor: Using `%u` for `unsigned long`. Should be `%lu` then. Coverity Scan just has some new checks. It is not the case that we introduced new bugs.

sriemer:
Coverity Scan reports four locations with printf format issues.
%u should be %lu or %d should be %lu. This issue has existed since
2009 with the 64 bit support introduction.
So fix the reported issues.

Fixes #351

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -640,7 +640,7 @@ bool handler__dregion(globals_t *vars, char **argv, unsigned argc)
 
         /* check if a match was found */
         if (np == NULL) {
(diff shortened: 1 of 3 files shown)


sriemer:
Travis-ci deprecated Trusty now as well. This is why the test is failing. Will be handled in another PR. Already approved by @12345ieee. So merging.

Issue #355, opened 2019-10-13
GC Question: How to use labels like gamename.exe+0efb


DHLF:
I am aware that games (except very old dos games) moves around in the computers memory every time you start a game. 
Therefore already found valid adresses would be of no use the next time.
Sadly GC does only allow me to change the description, type(int/float/string) and value of an entry in the cheatlist-window EXCEPT of the adress field.

To avoid typing in plenty(already) found adresses over and over again I am using bash scripts for any game which asks the current (base) adress and writes a table adding relative values.
This way I just have to find the current base adress and load the so created table back in GC.

Is it possible to use **relative labels** in the adding adress field(ctrl+M) like NameOfProcess+Value? (like it was possible in CE..)
If so I have not figured out how to do this YET.

While running dosbox I tried:
[dosbox] or {dosbox} or "dosbox" in the adress field, but neither was valid.
The process id was also of no use.
Also the use of +12345(as relative addition) was not possible.

It seems to me as if this is not implemented yet.

----
GC is a very great help for me to.. make some games easier ;)
Without it.. i would have been stuck in the windows world much longer..

sriemer:
Indeed, PIE (position-independent executable) support is not implemented. If GC would have PIE support, somebody would ask for PIC (position-independent code, static memory in libraries) support and then for pointer following. Then you'll notice that pointer following on Linux is not very practical as it is too hard to adapt to version, compiler, or distro updates and does not work well if you have multiple objects in a vector/list. Then you want to have API hooking and then you hit a design flaw in scanmem/GC as it is not able to run games/apps. Scanmem and GC don't claim to be game trainers.

If you can live without a GUI in favor of really working game trainer features, then [ugtrain](https://github.com/ugtrain/ugtrain) is the tool you need. `scanmem` has support for PIE/PIC static memory discovery as it provides match offsets and ELF load addresses. Forward stack offset support is also queued in a PR. That data can be put into `ugtrain` configs and `ugtrain` detects PIE automatically.

asdf23:
Here is what I do..
Turn off memory randomization
Save the state of the rom disk image
Change game state
Save the state of the rom disk image
Use scanmem to find memory location
Use script to accept memory location to return a set of scanmem commands..
Appended to with & #! important
Remap joystick button to script execution command

DHLF:
Maybe.. this could also be automated by scripting in a different way:

Lets try "Might&Magic VI",
[...]

PR #362, opened 2019-11-20
show_message: check for terminal connection before getting pager

bkazemi:
When scanmem is connected to a pipe or redirected to a file, stdout
is no longer associated with the terminal. This causes unnecessary
warning messages to appear. Add checks to avoid this.

handlers.c:
* Minor logic changes to pager code in handler__list().

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -327,8 +327,7 @@ bool handler__list(globals_t *vars, char **argv, unsigned argc)
     char *v = NULL;
     const char *bytearray_suffix = ", [bytearray]";
     const char *string_suffix = ", [string]";
-    struct winsize w;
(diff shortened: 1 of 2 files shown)


12345ieee:
While I'm not an expert in this kind of stuff, it looks good to me, and it works on all the cases I could think of.

Thanks, merging as-is, fixes #355 .

DHLF:
> While I'm not an expert in this kind of stuff, it looks good to me, and it works on all the cases I could think of.
> 
> Thanks, merging as-is, fixes #355 .

Thank you for your work!
Those error message has disappeared and I am still able to control scanmem using scripts.

PR #363, opened 2019-11-23
handlers: fix bugs in help command

bkazemi:
First bug fixed is not allowing the user to supply more than one
argument. The second bug fixed is rearranging the code to report
errors before opening up the pager. Without this, the pager would
open with no output.


12345ieee:
This has sat here for so long, sorry.
I remember testing it and finding no problems.

I can merge it if @sriemer has nothing to add.

12345ieee:
Merged in 46c257a thanks.
Closing.

PR #369, opened 2020-03-27
maps: Replace alloca() with variable length array

### LINKED ISSUE #368: scanmem crashes due to out of stack memory for filename with huge maps file
I seem to be having a problem where scanmem won't attach to a process which has more than 4GB allocated resident memory.  Specifically, I've been trying to attach it to a Borderlands 3 process (running via Wine) -- on the main menu and in some maps, the resident memory usage will remain below 4GB, but on other maps (and after moving around the previously-okay maps for a bit, causing more data to be loaded into RAM), it'll start refusing to connect, dying on startup with:

    info: maps file located at /proc/2897523/maps opened.
    Segmentation fault (core dumped)

If I hop back to the main menu, the resident memory usage will drop below 4GB again and I'll be able to connect, but some of the data I'm looking for is only accessible while ingame.

It's possible, of course, that the RAM thing is a red herring and there's something else going on, but that's what it seems to me.  This is probably the first >4GB process I've tried connecting to, so I don't have any data points from previous attempts to go on.

I've tried this both with the released scanmem-0.17 and with the current github HEAD, both with the same results.  Let me know if I can provide any more info!

sriemer:
When calling alloca() for the file name in the getline() loop, then
this allocates more and more memory on the stack for all file names
within the same maps file. It has no clue about the scope/block for
which it should keep the memory. It keeps it for the whole
sm_readmaps() function. With too many memory regions, this crashes.

So replace this with a variable length array as this one respects
the scope/block it is running in, restores the stack pointer at the
start of each loop cycle and only keeps one file name on the stack
at once this way.

Fixes #368

### DIFF
--- a/maps.c
+++ b/maps.c
@@ -31,7 +31,6 @@
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
-#include <alloca.h>
 #include <stdbool.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


sriemer:
Just FYI. Already approved. Merging.

Issue #228, opened 2016-10-27
Hexview looks weird in Gtk 3.22


12345ieee:
A new Gtk release, new problems.

I solved a bunch of 3.22-related things, this is the only one remaining.

12345ieee:
Figured where the problem is, it isn't Gtk 3.22, it's the text scaling factor I use.
Even though the hexview font doesn't change, once the factor goes over 1.08 the whole layout snaps.

Probably related to how the hexview widgets decide to dimension themselves.

12345ieee:
Yep, related to how the window is dimensioned.
Sadly I don't know enough about custom widgets to fix this.

Tagging as "Enhancement" (because it only happens for non-standard text scaling) and "help wanted", in case someone wants to have a look at the hexview.

PR #374, opened 2020-04-11
fixed broken HexView dispaly in none Latin locale system

shenada:
346de29..7cb8662
Those two commits fixed broken HexView dispaly in none Latin locale system.
Tested on Fedora 32 x86_64 LANG=[C, en_US, de_DE, en_UK, it_IT, zh_TW, zh_HK, ja_JP].

c9af3b7
Make sure test program also follow same AM_CFLAGS and take extra CFLAGS from ./configure CFLAGS=xxx
P.S: I got 4%~6% faster scan ./configure with CFLAGS=-O3 -g -march=native then CFLAGS=-O2 -g by compile with gcc 10 :smile:

Please be kind to review those changes.

### DIFF
--- a/gui/hexview.py
+++ b/gui/hexview.py
@@ -41,6 +41,7 @@ class BaseText(Gtk.TextView):
         self.override_font(Pango.FontDescription(parent.font))
         self.set_editable(False)
         self.texttag = self.buffer.create_tag(None)
+        self.texttag.set_property('fallback', False)
 
 GObject.type_register(BaseText)
 
--- a/gui/hexview.py
+++ b/gui/hexview.py
@@ -92,8 +92,11 @@ class OffsetText(BaseText):
         ctx = self.get_pango_context()
         font = ctx.load_font(Pango.FontDescription(self._parent.font))
         metric = font.get_metrics(ctx.get_language())
+        char_width_pango = max(
+            metric.get_approximate_char_width(),
+            metric.get_approximate_digit_width()) / Pango.SCALE
 
-        w = metric.get_approximate_char_width() / Pango.SCALE * (self.off_len + 1)
+        w = char_width_pango * self.off_len
         w += 2
 
         return w,w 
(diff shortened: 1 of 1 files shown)


12345ieee:
About the GUI font width miscalculation, what exactly is the issue?
I.e. can you post a screen of the current version in a locale that breaks?

shenada:
Before patch:
Screenshot from 2020-04-15 10-37-04

After patch:
Screenshot from 2020-04-15 10-39-59

I believe this is either freetype or GTK Pango layout bug.

shenada:
Here are three options to to fix this issue(currently know and tested):
1. set fallback to False. (Use the font which matched in current locale's Monospace font family).
2. set GUI locale to C or en_US.
3. set HexView font to any mono space(fixed pitch) font instead of Monospace font family.

12345ieee:
On my machine the issues aren't that bad, but indeed I see a messed-up layout with the `zh_TW.UTF-8` locale, and your patch fixes it.

I also can't replicate #228 anymore in any locale either before or after your changes, I'm not sure if it was related, but that's gone, too.

I honestly can't judge what you changed, tell me if that is the best way to fix this, if it is I'll commit it, as it improves things.

About the `CFLAGS` commit, I prefer having them explicitly written down, thanks anyway.

shenada:
> On my machine the issues aren't that bad, but indeed I see a messed-up layout with the `zh_TW.UTF-8` locale, and your patch fixes it.
> 
> I also can't replicate #228 anymore in any locale either before or after your changes, I'm not sure if it was related, but that's gone, too.
About #228. I can't reproduce on Fedora 29-32 too. May be it's about size rounding in Pango or about differences between python2/3(I abandon python 2 for a long time) ?

> I honestly can't judge what you changed, tell me if that is the best way to fix this, if it is I'll commit it, as it improves things.
Yes, I believe set fallback to False and calculate maxium approximate width is the best workaround for now. If we intend to do localization someday in the future.

> About the `CFLAGS` commit, I prefer having them explicitly written down, thanks anyway.
Understood. From my point of view, explicit CFLAGS usually means exception in build system. But I'll respect your preference without problem.
Please ignore c9af3b7.

shenada:
> I honestly can't judge what you changed, tell me if that is the best way to fix this, if it is I'll commit it, as it improves things.

Yes, I believe set fallback to False and calculate maxium approximate width is the best workaround for now. If we intend to do localization someday in the future.

> About the CFLAGS commit, I prefer having them explicitly written down, thanks anyway.

c9af3b7 removed and rebase to b480be48780842199dd69540ec3e75a8d1457bf9.

12345ieee:
Merged, thanks.

PR #377, opened 2020-04-20
Fix warning when compile with -Wextra flags and fix for potential bug.

shenada:
Fix warning when compile with -Wextra flags.

Including fix for potential bug cause by  sigsetjmp, siglongjmp in interrupt.c and interrupt.h.

Unsigned counter(i, j) also helps compiler gen hardware instructions to prevent underflow to negative value when needed.

np may become o or negative value from snprintf when out of memory. But we already check <=0 case. Simply cast to size_t then comparing is safe.

### DIFF
--- a/scanmem.c
+++ b/scanmem.c
@@ -64,9 +64,10 @@ globals_t sm_globals = {
         0,                      /* debug */
         0,                      /* backend */
         ANYINTEGER,             /* scan_data_type */
-        REGION_HEAP_STACK_EXECUTABLE_BSS, /* region_detail_level */ 
+        REGION_HEAP_STACK_EXECUTABLE_BSS, /* region_detail_level */
         1,                      /* dump_with_ascii */
         0,                      /* reverse_endianness */
+        0,                      /* no_ptrace */
     }
 };
 
--- a/targetmem.h
+++ b/targetmem.h
@@ -258,7 +258,7 @@ data_to_val_aux (const matches_and_old_values_swath *swath,
                  size_t index, size_t swath_length)
 {
     unsigned int i;
-    value_t val;
+    volatile value_t val;
     size_t max_bytes = swath_length - index;
 
     /* Init all possible flags in a single go.
--- a/handlers.c
+++ b/handlers.c
@@ -1271,7 +1271,7 @@ bool handler__dump(globals_t * vars, char **argv, unsigned argc)
     void *addr;
     char *endptr;
     char *buf = NULL;
-    int len;
+    unsigned len;
     bool dump_to_file = false;
     FILE *dump_f = NULL;
 
@@ -1349,7 +1349,7 @@ bool handler__dump(globals_t * vars, char **argv, unsigned argc)
         else
         {
             /* print it out nicely */
-            int i,j;
+            unsigned i,j;
             int buf_idx = 0;
             for (i = 0; i + 16 < len; i += 16)
             {
--- a/handlers.c
+++ b/handlers.c
@@ -90,9 +90,10 @@
 
 bool handler__set(globals_t * vars, char **argv, unsigned argc)
 {
-    unsigned block, seconds = 1;
+    unsigned block;
+	volatile unsigned seconds = 1;
     char *delay = NULL;
-    bool cont = false;
+    volatile bool cont = false;
     struct setting {
         char *matchids;
         char *value;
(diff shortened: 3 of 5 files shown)


scanmem.c
```diff
@@ -67,6 +67,7 @@ globals_t sm_globals = {
         REGION_HEAP_STACK_EXECUTABLE_BSS, /* region_detail_level */ 
         1,                      /* dump_with_ascii */
         0,                      /* reverse_endianness */
+        0,                      /* no_pthread */
```
12345ieee, line 70: The comment is wrong here, should be `no_ptrace`.

shenada: > The comment is wrong here, should be `no_ptrace`.

Done.  Rebased to 76b24ca2cae5748d7dd34f1eac30e19b632b3142.

shenada: Done. Rebased to 76b24ca.

shenada: Done. Rebased to 76b24ca.

12345ieee:
I've merged the initializers commit, I'll take some more time to review the other 2.
Thanks.

shenada:
> I've merged the initializers commit, I'll take some more time to review the other 2.

Sure.

targetmem.h
```diff
@@ -258,7 +258,7 @@ data_to_val_aux (const matches_and_old_values_swath *swath,
                  size_t index, size_t swath_length)
 {
     unsigned int i;
-    value_t val;
+    volatile value_t val;
```
12345ieee, line 261: I don't see this one when using `-Wclobbered`, how do you get it?

shenada:
GCC 10 + -Wextra
It's a new compiler feature provides strong static analysis.

12345ieee:
Yeah, that's what I used: `make CC=gcc-10 CFLAGS='-O2 -Wextra -Wno-unused-parameter' -j`.
In any case, I'd prefer to understand why they are clobbered and try to fix that instead of slapping a `volatile` on them.

The unsigned one is probably easier to check and validate, I'll check it tomorrow or so.

shenada:
> Yeah, that's what I used: `make CC=gcc-10 CFLAGS='-O2 -Wextra -Wno-unused-parameter' -j`.
> In any case, I'd prefer to understand why they are clobbered and try to fix that instead of slapping a `volatile` on them.
> 
> The unsigned one is probably easier to check and validate, I'll check it tomorrow or so.

For two reason,
1. signal could interrupt any time.
intr_used must guard whole signal handler in every handler.
Or atomic/volatile access every variable across handler guardian or different handler.

`unsigned seconds = 1;` and `seconds++;` 
`bool cont = false;` and `cont = true;`
access across guardian.

2. unsigned intr_used is not safe.
intr_used must atomic or volatile. Could be both atomic and volatile.
like `volatile sig_atomic_t intr_used`
or access through [GCC Atomic-Builtins](https://gcc.gnu.org/onlinedocs/gcc-4.1.1/gcc/Atomic-Builtins.html)
or C11 stdatomic.h

e.g:
Signal 1st launched and enter handler 1st ->
Before CPU reach or complete `intr_used=1` instructions ->
At the same time. Signal 2nd launched and enter handler 2nd

in this case intr_used could be 0(case before reach) or uncertain 0/1(case during set instructions complete or because CPU superscaling ... etc)

targetmem.h
shenada: Because value_t val is a struct type.
We don't know how compiler return/pass struct after optimization.
May through stack, register or memory access. Which depends on compiler decide really inline this function or not, and the usage of register.

According C standard, inline keyword for function which only recommend compiler inline the function. Not guaranteed compiler always inline the function.
Unless use this syntax `inline __attribute__((always_inline))`

In the other hand, pass/return struct usually gen more then one instruction. Once interrupt between instructions, struct data may incomplete which results unpredictable behavior.

[GCC Aggregate Return Impl](https://gcc.gnu.org/onlinedocs/gcc-3.3.6/gccint/Aggregate-Return.html)
In C99 the implantation of inline not defined very strict. And also GCC in different version or optimization level may have different behavior. 
[Inline Functions In C](https://www.greenend.org.uk/rjk/tech/inline.html)

So, basically return/pass struct in inline function without atomic access is a bad idea.
`volatile` here basically tells compiler not optimize and try its best to keep access atomically.

shenada:
Here is another way to fix clobbered access 2a9d66a0fe4d903ee4c835a3b3f238292ec6789b

PR #380, opened 2020-05-05
cleanup: Remove unwanted tail space

shenada:
Simply remove unwanted tail space for eye candy

Signed-off-by: Dark Shenada <shenada@gmail.com>


12345ieee:
While I like the changes in themselves, and you're welcome to apply them in a file if you're doing another change in it, I don't want to commit a mass whitespace manipulation like this.

Thank you for the offer, but I prefer to reject this.

PR #381, opened 2020-05-05
Split Wclobbered fixup from pull request #377

shenada:
https://github.com/scanmem/scanmem/pull/377#issuecomment-623762128
https://github.com/scanmem/scanmem/pull/377#discussion_r420419049
https://github.com/scanmem/scanmem/pull/377#issuecomment-623849478
https://github.com/scanmem/scanmem/pull/377#issuecomment-624323867

Some variable might be clobbered by sigsetjmp, siglongjmp in interrupt .c and .h.
Which means value in register or stack may not safe in some aggressive optimization.

Another way to fix Wextra warnings: Wclobbered without using gcc attribute

Signed-off-by: Dark Shenada <shenada@gmail.com>

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -90,9 +90,9 @@
 
 bool handler__set(globals_t * vars, char **argv, unsigned argc)
 {
-    unsigned block, seconds = 1;
+    unsigned block, seconds;
     char *delay = NULL;
-    bool cont = false;
+    bool cont;
     struct setting {
         char *matchids;
         char *value;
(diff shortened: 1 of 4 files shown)


1rhino2:
Confirmed this is folded into #457 as-is (`57cfe40` for the handlers.c piece, `1d8553f`/`8f01f89` for the interrupt.c/interrupt.h/targetmem.h volatile pieces). Diffed both against your original patch and they match. Thanks for catching these, closing this one out once #457 lands.

1rhino2:
the Wclobbered fix from this went in with #457, merged as 2341d8e. thanks shenada. closing since it is on main now.

PR #383, opened 2020-05-23
Completed Russian translation.

MrGooogo:
Filled in the gaps, fixed a few mistakes from previous version of translation(some typos and misplacement of symbols), added keyboard shortcuts tips that were cutted out of the translation previously(for no apparent reason), added the correct plural-dorm formula. All in all there were only quite minor changes.

While i were at it, i found some strange entries. To be exact i found two of them :  `<unknown>` and `Addres_s:` . First one seems like a tag of sorts, so i am unsure if it was supposed to be translated, but i translated it anyway. Second one seems like a typo on  an original text part(or so it seems from the context provided in gameconqueror.ui) so i just translated it normally without any more thought given? please correct if i am wrong though.


12345ieee:
Thanks for the contribution, answers:
* we don't normally have a Russian translator, so when new strings are introduced, they are left untranslated in Russian
* it's fine to translate `<unknown>`
* `Addres_s` and similar isn't a typo, it's a notation that tells GTK about an hotkey association. If you still have the character after the `_` in the translated word, leave the `_`. Otherwise remove it, as you've done for that word.

PR #389, opened 2020-07-27
added undo feature

vikke1234:
you can set the undo limit with the command `option undo_limit N`, you
can also redo undone scans

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -542,6 +542,15 @@ bool handler__reset(globals_t * vars, char **argv, unsigned argc)
         return false;
     }
 
+    /* reset history */
+    struct history_entry_t *cur = NULL, *next = NULL;
+    cur = TAILQ_FIRST(&vars->match_history);
+    while(cur != NULL) {
+        next = TAILQ_NEXT(cur, list);
+        free(cur);
+        cur = next;
+    }
+    TAILQ_INIT(&vars->match_history);
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


1rhino2:
Like the feature, undo/redo for scan rounds is genuinely useful. Found a real memory-safety bug testing it though, reproduced cleanly under ASan.

`handler__reset` frees every entry in `vars->match_history` and does `TAILQ_INIT()` to empty the list, but never sets `vars->current = NULL`. `handler__redo` guards on `vars->current == NULL`, not on the list being empty, so after a `reset` it still dereferences the now-freed `vars->current` via `TAILQ_NEXT(vars->current, list)`. Straight use-after-free.

Repro (`option undo_limit 5;option scan_data_type int8;1;1;reset;redo`) under `-fsanitize=address`:

```
==ERROR: AddressSanitizer: heap-use-after-free on address 0x7be37d1e0b80
READ of size 8 at 0x7be37d1e0b80 thread T0
    #0 handler__redo handlers.c:1799
    #1 sm_execcommand commands.c:272
freed by thread T0 here:
    #0 free
    #1 handler__reset handlers.c:550
previously allocated by thread T0 here:
    #0 malloc
    #1 handler__default handlers.c:1017
```

Fix is small, `handler__reset` just needs `vars->current = NULL;` alongside the `TAILQ_INIT`, and `handler__undo`/`handler__redo` should probably check `TAILQ_EMPTY()` (like `handler__undo` already does at the top) instead of `vars->current == NULL`, so they don't depend on `reset` remembering to clear it.

Two smaller things while I was reading the history-push code in `handler__default`:
- `entry = malloc(sizeof(struct history_entry_t))` has no NULL check before `entry->num_matches = ...` a few lines down. Narrow OOM-only path but it's an unchecked deref.
- `sm_undo_scan()` / `sm_redo_scan()` in scanmem.c are dead stubs that just `return true;` and aren't called from anywhere, the real logic lives inline in `handler__undo`/`handler__redo`. If these are meant to be the library-facing API, they should probably call into the real logic instead of no-op'ing, otherwise worth dropping them so nobody calls them expecting an actual undo.

Happy to help land the `vars->current` fix if you want to keep pushing this one, everything else about the approach (TAILQ-based history, truncating the redo branch on a new scan after undo) looks right.

PR #394, opened 2020-12-16
Add support to ppc64le focal

gururajrkatti:
This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,12 @@ language: c
 sudo: required
 dist: bionic
 
+jobs:
+  include:
+    - arch: amd64
+      dist: bionic
+    - arch: ppc64le
+      dist: focal   
 env:
   global:
   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created


12345ieee:
I don't see great value in testing another 64bit little endian architecture.
If travis had 32bit arches or big endian arches, it'd be more interesting.

Are you involved in packaging scanmem on Ubuntu?

gururajrkatti:
Thanks @12345ieee  for the response. I agree that both ppc64le and amd64 are little endian. However from our previous experience we have seen porting issues/arch differences even between them such as data types, compiler flags, assembly/low level code. Page size is different between them. There can be missing dependencies. This PR will ensure to catch any such issues on a continuous basis.

12345ieee:
I've read up on the arches supported on travis, and they're all 64bit little endian arches.
I see little point burdening travis further and slowing the CI checks to test again on the same conditions.

Thanks for the interest and the PR, but I have to refuse it.

PR #395, opened 2021-01-01
allow configure.ac to be cross-friendly

ndowens:
Use --enable/--disable to determine if CHECK_FILES should be ran

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -59,7 +59,16 @@ case "$host_os" in
     ;;
 esac
 
+# Allows disabling procmem support
+AC_ARG_ENABLE(procmem, [AS_HELP_STRING([--disable-procmem],
+                         [forcefully disable proc/pid/mem support])])
+AS_IF([test "x$enable_procmem" = "xno"], [
+  AC_DEFINE(HAVE_PROCMEM, [0], [Enable /proc/pid/mem support])
+])
+
 AS_IF([test "x$android" = "xno"], [
+  AS_IF([test "x$enable_procmem" != "xno"],
+		[
   # also need to check if the file is zero'ed (some hardened systems)
   AC_CHECK_FILE([/proc/self/maps], [], [
     echo "This system does not seem to have /proc/pid/maps files."
@@ -76,7 +85,7 @@ AS_IF([test "x$android" = "xno"], [
     echo "This system does not seem to have /proc/pid/mem files."
     echo "Falling back to ptrace() only support."
     AC_DEFINE(HAVE_PROCMEM, [0], [Enable /proc/pid/mem support])
-  ])
+  ])])
   # malloc optimizations without Android
   AC_FUNC_MALLOC
   AC_FUNC_REALLOC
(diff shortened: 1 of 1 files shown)


ndowens:
https://github.com/void-linux/void-packages/pull/27590 shows success after using this patch, the one error currently shown at time of writing this is just a make check test failing as musl is a little different, nothing to do with build.

12345ieee:
I remember cross-compile failing, but did not expect this to be the only issue.

If it really is, I'll happily merge.

ndowens:
> I remember cross-compile failing, but did not expect this to be the only issue.
> 
> If it really is, I'll happily merge.

https://github.com/void-linux/void-packages/pull/27590 You can look and see it did cross build with this simple fix

Johnnynator:
This isn't correct and can be potentially hurtfull, you do check if these file exist in the build environment system, but we care about the target system. There is to my knowledge now standard way for anything like this in autoconf, so you would need to add a configure flag for this that is needed for cross builds (and can optionally also used for non cross, e.g. if the build environment might differ to the target).

There is already `--disable-procmem` defined, so the check if this (or `--enable-procme`) is set, should probably just be moved before the check, if procmen exists, and skip the `AC_CHECK_FILE`.

And the `/proc/self/maps` maybe as an optional warning or skip it when cross compiling.

12345ieee:
You are correct, I would expect someone cross-compiling to set `--{enable,disable}-procmem` correctly and then the build to ignore the actual procmem check.

I know the PR is a workaround, ideally we'd have:
```
if not cross-compiling:
    AC_CHECK_FILE
else
    believe the user's choice and check nothing
```

if such if can be written it would be better than the PR.

ndowens:
I'll see what I can come up with

ndowens:
Ok I've added an --enable-procmem switch and added another test check to see if it should run procmem test according to test result of enable/disable. Currently it is enabled by default, but doing this allows cross-building too, if --disable-procmem is used

ndowens:
Ok first I am an autotools noob so keep that in mind :)
I've updated with a patch that eliminates check_file and use --enable/--disable-procmem, which is set off default in case 
a user's system doesn't support it. 

If I didn't mis-interpret the original configure.ac, I intended it to be along the lines of:
if android = no 
if enable-procmem = yes ; then 
    AC_SYS_LARGEFILE
    Define HAVE_PROCMEM=1
else(enable-procmem):
   AC_FUNC_MALLOC
   AC_FUNC_REALLOC
else(android)
Define HAVE_PROCMEM=0

I think I wrote it like I understood original to essentially match what I believe is happening and intended.

ndowens:
Moved the procmen arg declaration, currently it is enabled by default, but if someone needs to cross-compile, it atleast allows procmem to be disabled and skip checks that causes cross fail.

12345ieee:
I've verified via:
```sh
./configure --{dis,en}able-procmem --host=arm-linux-{gnu,android}eabi
```
that the flag is set correctly each time.

Merged in 5d2b7fc , thanks.

PR #396, opened 2021-01-01
test/memfake.c: use unsigned int  instead of uint

ndowens:
musl test gave uint unknown variable. Using int fixed test

### DIFF
--- a/test/memfake.c
+++ b/test/memfake.c
@@ -25,7 +25,7 @@
 
 int main(int argc, char **argv)
 {
-    uint MB_to_allocate = 1;
+    unsigned int MB_to_allocate = 1;
     bool add_randomness = false;
 
     if (argc >= 2) MB_to_allocate = atoi(argv[1]);


12345ieee:
Oh yeah, libc has a bunch of nice typedefs that aren't really standard.

I'd rather it became `unsigned int` instead, then I'll merge.

ndowens:
> Oh yeah, libc has a bunch of nice typedefs that aren't really standard.
> 
> I'd rather it became `unsigned int` instead, then I'll merge.

Fixed

shenada:
On Fri, Jan 1, 2021, 23:47 Andrea Stacchiotti <notifications@github.com> wrote:
>
> Oh yeah, libc has a bunch of nice typedefs that aren't really standard.
>
> I'd rather it became unsigned int instead, then I'll merge.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.

Best choice is using size_t for MB_to_allocate.
And use strtoul() instead of atoi() because atoi() returns int, even
worse if the value can not be represented, behavior is undefined.

patch example as below:
=====================================================================
diff --git a/test/memfake.c b/test/memfake.c
index 1c4fa94..b619840 100644
--- a/test/memfake.c
+++ b/test/memfake.c
@@ -22,19 +22,21 @@
 #include <stdlib.h>
 #include <time.h>
 #include <unistd.h>
+#include <assert.h>

 int main(int argc, char **argv)
 {
-    uint MB_to_allocate = 1;
+    size_t MB_to_allocate = 1;
     bool add_randomness = false;

-    if (argc >= 2) MB_to_allocate = atoi(argv[1]);
-    if (argc >= 3) add_randomness = atoi(argv[2]);
+    if (argc >= 2) MB_to_allocate = strtoul(argv[1], 0, 10);
+    if (argc >= 3) add_randomness = strtoul(argv[2], 0, 10);
     if (argc >= 4) return 1;

     size_t array_size = MB_to_allocate * 1024 * 1024 / sizeof(int);

     int* array = calloc(array_size, sizeof(int));
+    assert(array != NULL);

     // Fill half with random values and leave an half of zeroes, if asked to
     if (add_randomness) {
=====================================================================

12345ieee:
Merged @shenada 's version, thanks.

PR #403, opened 2021-08-09
Added real time memory region monitoring feature.

vandrlexay:
New "memdiff" command that scans given memory fragment and displays
all changed values.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -1241,6 +1241,204 @@ bool handler__watch(globals_t * vars, char **argv, unsigned argc)
     }
 }
 
+bool handler__memdiff(globals_t * vars, char **argv, unsigned argc)
+{
+    void *addr;
+    char *endptr;
+    char *buf = NULL;
+    char *prev_buf = NULL;
+    char *start_buf = NULL;
+    char use_list_output = 0;
+    size_t len;
+
+    if (argc < 3 || argc > 4)
+    {
+        show_error("bad argument, see `help memdiff`.\n");
+        return false;
+    }
+
+    /* check address */
+    errno = 0;
+    addr = (void *)(strtoll(argv[1], &endptr, 16));
+    if ((errno != 0) || (*endptr != '\0'))
+    {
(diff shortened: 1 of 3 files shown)


1rhino2:
Tested against known ground truth instead of just reading the diff. Wrote 16 distinct bytes (`0x10`..`0x1f`) into a target process and confirmed them with the existing `dump` command:

```
> dump 2008fff7000 16
0x2008fff7000: 10 11 12 13 14 15 16 17 18 19 1A 1B 1C 1D 1E 1F
```

Then ran `memdiff` on the exact same address/length:

```
> memdiff 2008fff7000 16
0x2008fff7000: 11 12 13 14 15 16 17 18 19 1A 1B 1C 1D 1E 1F BE
```

Off by one: byte 0 (`0x10`) never shows up at all, every real byte after it is shifted one position left, and the trailing `BE` is uninitialized heap memory, not part of the actual 16-byte region.

Root cause is in the table-mode print loop, `buf_idx` starts at 0 but gets pre-incremented (`buf_idx++`) before the first `buf[buf_idx]` read in each row, so the very first access is `buf[1]`, not `buf[0]`. Same `buf_idx` variable then carries the same off-by-one into the ascii column's indexing too, except the ascii column separately uses `buf[i+j]` (not `buf_idx`), so hex and ascii ends up showing different offsets on the same line when `dump_with_ascii` is on. And since `buf`/`prev_buf`/`start_buf` are `malloc(len + sizeof(long))` without the extra padding ever being written, the shifted-forward read walks into that uninitialized tail on the last byte of each row, which is where the `BE` came from, contents of whatever was previously on that heap page.

List mode (`memdiff addr len list`) doesn't have this, it correctly uses `buf[i]`.

Fix is to just use `buf[i + j]` (post-increment or plain index math) consistently in both the hex and ascii table-mode loops instead of a separately-tracked `buf_idx`, same as list mode already does. Feature itself is a nice one to have (live diff view with a first-vs-last-change color split), worth fixing up.

PR #408, opened 2022-02-21
support other privilege execution utilities

petteyg:
Update the launcher script for the GUI with support for other privilege escalation tools that may use sudo instead of requiring root password entry.

### DIFF
--- a/gui/gameconqueror.in
+++ b/gui/gameconqueror.in
@@ -1,10 +1,20 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 DATADIR=@PKGDATADIR@
 
-PKEXEC=$(command -v "pkexec")
-if [ -n "$PKEXEC" ]; then
-    $PKEXEC $DATADIR/GameConqueror.py "$@"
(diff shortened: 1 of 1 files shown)


12345ieee:
It used to work exactly like you've put it, but was changed to `pkexec`-only in https://github.com/scanmem/scanmem/commit/b75f1be2768d983097add9c2676396085f97a2c1 .

Is there some distro where the current script fails?

petteyg:
Not that it fails, but that I do not want to enter the root password when it is entirely unnecessary to do so. If a tool is installed that allows using a backend such as sudo instead where I do not need to know the root password, it should use that tool.

12345ieee:
If that's what you want to achieve, just invoke `$privilege_escalation_utility gameconqueror` to delegate the authentication to said utility.
No need to list every privilege escalation utility in the launch script.

petteyg:
Then why is there a launch script at all, if you're just going to leave it to the distro or every individual user to sanitize, instead of making a best-guess as to what tool they might prefer?

Issue #406, opened 2021-10-08
[Feature Request] Allow libscanmem to search for executable regions


korcankaraokcu:
Libscanmem currently discards executable regions even when scanning with full scope. It can be quite useful for debugger integrations, such as PINCE. This feature is also requested in https://github.com/korcankaraokcu/PINCE/issues/107

sriemer:
Could be easily done by adding a scan level 4 I guess. Got no time to implement this unfortunately. Feel free to contribute.

pollev:
Hello,

I also need this feature (and would also like to see it added to PINCE).
I have opened a PR for this: https://github.com/scanmem/scanmem/pull/410
Please give feedback and merge if appropriate.

PR #410, opened 2022-04-21
enable searching in read-only memory (PR for issue #406)

pollev:
Hi,

I also ran into issue https://github.com/scanmem/scanmem/issues/406
This functionality is important for scanning executable regions.
Specifically for AOB scanning in game hacking this is an incredibly useful feature.
Merging this should also enable this feature for PINCE (https://github.com/korcankaraokcu/PINCE)

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -1651,7 +1651,8 @@ bool handler__option(globals_t * vars, char **argv, unsigned argc)
     {
         if (strcmp(argv[2], "1") == 0) {vars->options.region_scan_level = REGION_HEAP_STACK_EXECUTABLE; }
         else if (strcmp(argv[2], "2") == 0) {vars->options.region_scan_level = REGION_HEAP_STACK_EXECUTABLE_BSS; }
-        else if (strcmp(argv[2], "3") == 0) {vars->options.region_scan_level = REGION_ALL; }
+        else if (strcmp(argv[2], "3") == 0) {vars->options.region_scan_level = REGION_ALL_RW; }
+        else if (strcmp(argv[2], "4") == 0) {vars->options.region_scan_level = REGION_ALL; }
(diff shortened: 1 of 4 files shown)


eLeCtrOssSnake:
I quick-checked the code and it seems to be legit but currently, I cannot test it. This is a feature that I've requested back in PINCE too so I would like to see it merged.

sriemer:
Looks like the patch would work but there is too much duplication of the writable region check and too complex if statements are created here.
Just add a check if not REGION_ALL and write != ‚w‘, then continue. That is one simple maintainable check and pretty close to the original implementation.

pollev:
@sriemer Thank you for the feedback. I have refactored the case switch. Please let me know if it seems acceptable now.

12345ieee:
You should also update the documentation and autocompletion at `handlers.h` in the `OPTION_...` macros.

The rest looks good.

sriemer:
This is what I also wanted to suggest. Thanks, Andrea. I spend almost no time at a PC recently.

pollev:
@sriemer @12345ieee 
Done :)

pollev:
Hi, Is there anything else you would like to see updated? Do I need to validate something else to allow this to be merged?

12345ieee:
No, it looks good, it just needs me to have time, likely this weekend.

12345ieee:
Added the new level to GC, merged in [ac32e5d](https://github.com/scanmem/scanmem/commit/ac32e5d7d7429ec2123c5ff63559bc9ad37b4f09).

pollev:
Hi,

Sorry to bother you, but I noticed you named the old option "readonly" in gameconquerer. But that is the opposite of what it actually is. The old option is "writable regions only"

PR #418, opened 2022-12-27
Move contents of README to README.md

Infinixius:
On GitHub, visiting `README` gives you a text version without Markdown, and visiting `README.md` doesn't work. This fixes both issues by just getting rid of the symbolic links and keeping the Markdown versions.


sriemer:
For `autotools`, a file `README` is mandatory. But yes, reversing the logic is the way to go. Move the contents to `README.md` and put a symlink `README` to it.

sriemer (changes requested):
Add the `README` to `README.md` symlinks and then its fine.

12345ieee (changes requested):
I agree, the README is written with markdown syntax, so it should have a `.md` extension, but I'd rather keep the README symlink as @sriemer said.

sriemer (changes requested):
Something is wrong with your master branch. README.md is still a symlink to git. See: https://github.com/Infinixius/scanmem 
If I clone your repo, then both files `README.md` don't exist and `README` is a broken link. You should better test your work.

I actually wanted to check if you haven't changed the contents via
`meld ./scanmen/README ../Infinixius-scanmem/README.md`

sriemer:
Also contributors should have a real name and a valid email address. Otherwise, we risk getting sued ourselves if contributions contain anything that has been contributed without permission with parts from a third party. I'll fix this with own commits.

sriemer:
Merged with PR 419. Thanks for reporting. :smiley:

PR #419, opened 2023-01-09
Move contents of README to README.md

sriemer:
Successor of #418


sriemer:
Tested. Everything looks good now. So merging.

12345ieee:
The should have really been squashed, but it's done now.

sriemer:
Some of it could have been squashed, yes. But otherwise, we could run into the same issue again with the broken symlinks if doing too much stuff in the same commit.

Issue #423, opened 2023-02-05
Syntax for search in XOR'd memory region?


dejj:
I have implemented searching in XOR'd memory regions.
XOR'd memory is a simple cheat prevention technique. I am currently testing the feature.
Which syntax do you prefer?

I propose the prefix caret (^), with 1 or 2 arguments.
```c
#define XOR_LONGDOC "usage: ^ n [m]\n" \
                "Takes the value of the previous round (a^x) and the current round (b^x)\n" \
                "If m is given, match all values where (a^x^b^x)==(n^m)\n" \
                "Otherwise match all values where (a^x^b^x)==n\n" \
                "You can use this if memory has been xor'd and you know both clear values."
```

Pro: caret matches the C-operator.
Con: caret is a deadkey and needs to be pressed twice.

Pro: prefix operator (^ x y) makes arguments easier to parse.
Con: infix operator (x ^ y) is more commonly used.

12345ieee:
Do you have a link explaining this XOR technique? I'd like to read about it, then evaluate if adding it.

dejj:
Thanks for looking into this :)

> link explaining this XOR "anti-cheat" technique
It's just obfuscation. I doubt anybody considers it worth a write-up. Prevents nothing more than scanning for plain values with scanmem and CheatEngine.
I've seen XOR obfuscation used in Unity games. I used the xor-scan for "The Captain" yesterday. (It was no time for caution.)

Find my draft implementation in the following PR: https://github.com/scanmem/scanmem/pull/424

gringus:
This specific technique is described at https://www.gamecheetah.org/blog/bypassing-xor-encryption-game-guardian/ (look for `XOR encryption in mobile games`). Keep in mind this relies on static encryption key which probably does not change during the course of the game.
I tried to use it in [Heroes of Might and Magic VI](https://en.wikipedia.org/wiki/Might_%26_Magic_Heroes_VI) and it turns out they use slightly different implementation there: encrypted value and encryption key are stored as 2 consecutive `int32` values and they change on every real value change.
Both above implementations can be tackled with changed/not changed searches, but process is long, boring and prone to errors. Having dedicated XOR searches significantly speeds up the process, makes it more targeted and much more reliable.

1rhino2:
same note as on #376, connecting this to the code. the implementation is #424 (dejj), open since 2023.

on the syntax question in the original post, caret prefix reads fine to me and does not collide with anything else in the command set, so no objection there.

[...]

PR #424, opened 2023-02-14
Add support for XOR between scan rounds

dejj:
Implementation for https://github.com/scanmem/scanmem/issues/423.
I added float, double support for completeness

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -705,29 +705,46 @@ bool handler__lregions(globals_t * vars, char **argv, unsigned argc)
 bool handler__operators(globals_t * vars, char **argv, unsigned argc)
 {
     uservalue_t val;
+    uservalue_t val2;
     scan_match_type_t m;
 
-    if (argc == 1)
+    printf("parsing %d arguments\n", argc);
+
+    switch (argc)
     {
-        zero_uservalue(&val);
-    }
-    else if (argc > 2)
-    {
-        show_error("too many values specified, see `help %s`", argv[0]);
-        return false;
-    }
-    else
-    {
-        if (!parse_uservalue_number(argv[1], &val)) {
-            show_error("bad value specified, see `help %s`", argv[0]);
-            return false;
+        case 3:
+        {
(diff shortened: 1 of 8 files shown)


1rhino2:
Like the idea (xoring across rounds against a known plaintext pair is a real technique), but this isn't ready as-is, three separate things:

1. `printf("parsing %d arguments\n", argc);` at the top of `handler__operators` fires on every single operator command, `=`, `<`, `>`, everything, not just `^`. Leftover debug line, needs to go.

2. `printf("found FLOAT xor match\n");` inside `DEFINE_FLOAT_OPERATIONBY_ROUTINE_FLIP` fires once per candidate match during a float scan. On a scan with thousands of matches that's thousands of prints per round. Same deal, debug leftover.

3. Bigger one: the float/double XOR in `xor_uservalue()` and the float match routine both do `(unsigned long)get_f64b(x)`, a numeric cast, truncates to an integer value, not a reinterpret of the bit pattern. Real XOR-in-memory happens on raw bytes, so this doesn't actually undo a bitwise XOR applied to a float/double, it just XORs the truncated integer parts of the values. Confirmed by hand: `(unsigned long)3.14` is `3`, not the IEEE-754 bit pattern of `3.14`. You'd want something like a union or `memcpy` into a same-width integer type to get the actual bits. Integer XOR (int8/16/32/64) doesn't have this problem, that half is fine.

4. Tried running the modified `test/memfake.c` under how the test harness actually launches it (background, stdin closed): the new `scanf` loop never blocks on EOF the way you'd expect, it free-spins reading immediately-available EOF and printfs every iteration. Killed it after 6 seconds, it had already written 476MB to stdout. `sm_test.sh` backgrounds `memfake` with no stdin attached, so this would hang or blow out disk on CI as written. The interactive pattern-picker probably needs to be a separate standalone test binary, not baked into the same `memfake` the automated suite depends on.

Happy to help rework the float bit-cast + split out the interactive memfake changes if you want to keep pushing this one, the actual XOR-scan feature is worth having.

PR #425, opened 2023-02-15
Use 'unsigned int' instead of unofficial alias 'uint'

thesamesam:
glibc headers have 'uint' typedef'd to 'unsigned int', but this isn't guaranteed to be avaialble. Change to 'unsigned int' which is equivalent to fix building on musl.

Bug: https://bugs.gentoo.org/854840

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -392,7 +392,7 @@ bool sm_checkmatches(globals_t *vars,
         match_flags checkflags;
 
         match_flags old_flags = reading_swath_index->data[reading_iterator].match_info;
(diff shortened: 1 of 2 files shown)


sriemer (approved):
LGTM. Thanks. :smiley:

12345ieee (approved):
Definitely, I've done the change in the years around the code I've touched, these are just the last few. Merging, as I see @sriemer agrees as well.

12345ieee:
The build breaks due to an error in `apt upgrade`, nothing that matters for scanmem itself, this is fine to merge.

thesamesam:
Thanks folks!

Issue #447, opened 2024-11-25
memory editor GUI improvments


sqrt0x:
First of all, I’m a big fan of the Scanmem project and want to thank you for this masterpiece.

I'm here to talk about this https://github.com/scanmem/scanmem/issues/357
this ticket was opened in 2019 and has not been closed since so I’m curious if there have been any updates.

Is there any built-in feature/plugin I'm unaware of, or must I change the code and do a pull request?

### Goals:
- ~~Display every 4 bytes with different colors~~
- Display  memory as float/int values instead of hexadecimal ones

sqrt0x:
I can make the code cleaner and submit this as a pull request <br><br>

![image](https://github.com/user-attachments/assets/789973ae-662d-4b4e-b584-4bd8cde8c2db)

sriemer:
Sounds good. Thanks for the PR. Usually I only use the CLI tool but stopped this in 2019 when my health collapsed. You can imagine that me and 12345ieee have different priorities now. We are looking for new maintainers.
Thanks so much for starting with thanks. 💓 We put so much effort, hard work, and our leisure/sleep time into this project.
But now it is time for us to pass this over. Would you be interested? TIA

12345ieee:
Definitely, @sriemer can you add a big note to the readme that we're looking for people to take over?
Maybe also open an issue we can link to.

I know the folks at PINCE ( @korcankaraokcu ) might be interested, also I see a couple forks with recent significant work, like @OpenA 's and @sivertac 's.

FatSquare:
Hello,  thanks for you offer of course I'll be more than happy and ready to take over and contribute to this project. I'm using scanmem daily and I don't want to see it going forever. Either way, I'm thankful for what you guys have done for this project so far. Thanks again

PR #448, opened 2024-11-25
Add select display type to scanmem memory viewer

sqrt0x:
I added a feature to scanmem that lets you specify the rendering type in the memory browser. <br>

Issue related https://github.com/scanmem/scanmem/issues/447

https://github.com/user-attachments/assets/edc53b67-7c69-4433-b8ec-7bdfa4907e53

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -141,6 +141,8 @@ class GameConqueror():
         self.memoryeditor_hexview.show_all()
         self.memoryeditor_address_entry = self.builder.get_object('MemoryEditor_Address_Entry')
         self.memoryeditor_hexview.connect('char-changed', self.memoryeditor_hexview_char_changed_cb)
+        self.editmemory_dialog = self.builder.get_object('EditMemoryDialog')
+        self.memoryeditor_value_type  = self.builder.get_object('DisplayType_ComboBoxText')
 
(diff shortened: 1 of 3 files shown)


sriemer:
@FatSquare You caught my attention. This looks like a really useful feature. Thanks. Not sure when I will get to reviewing it though. Keeping it on my list.

1rhino2:
Correcting my last comment here, I said I'd run this against memfake but hadn't actually driven the GUI yet when I wrote that, only read the diff. Went back and actually did it: built with `--enable-gui`, ran GameConqueror's `HexView` widget headless under Xvfb, fed it a real byte buffer and cycled `display_type` through byte/int32/int64/float32/float64.

Checked the int32 case by hand: bytes `00 01 02 03` as little-endian int32 rendered as `50462976`, which is `0x03020100`, correct. float32/float64 came back as sane finite values too, no crash, no exceptions across all 5 modes.

This landed in #457 (`1913f8e`) with a follow-up commit keeping the python3 shebang and pinning `gi.require_version('Gtk', '3.0')` so it doesn't break on newer PyGI. Thanks for the feature, credited in the 457 summary. Sorry for the mixup in the earlier comment.

1rhino2:
your display type selector went in with #457, merged just now as 2341d8e, credited in the commit. thanks sqrt0x. closing since it is on main now.

Issue #392, opened 2020-10-23
Process already attached by a tracer


allenaigit:
This isn't really a bug, but it's a very confusing situation when encountered and I think it's worth at least being mentioned here.

**To reproduce:** attach to a process (say process ID 1000) with gdb or strace, then try to attach to the same process with scanmem. scanmem will produce error "failed to attach" as a result.

**Suggestion:** check the process's status before attempting to attach, and return a clear message to let user know the reason of failed attachment is due to the process already being traced. e.g. check the return of `grep TracerPid /proc/$PID/status` and if it isn't 0, a process has already attached to it.

Why this was important to me: I was poking around TotalWar Three Kingdoms and noticed that GameConqueror always returns 0 address disregard of value to search for. So I tried again with scanmem directly and got the error "failed to attach". However since both the game process and GC are all launched by me (under my username) and that I have sudo access, it is really weird I'd run into any kind of permission issues. After quite some time digging around google I finally nailed the problem to that game process already has tracer attached. That attached thing is made by game publisher to pass signals between their "launcher" and the game itself, so I killed the attached process and now I have full access to the game's memory again.

When I first launch the game, this is what I get with scanmem: 
```
error: failed to attach to 79436, Operation not permitted
error: failed to search target address space.
```

This is because:
```
root:~# grep TracerPid /proc/79436/status
TracerPid:	79435
```

12345ieee:
Yes, a program tracing itself is a common defense against memory editors on linux.

I welcome a PR that adds this feature to scanmem, but I can't promise I'll work on it myself.

PR #455, opened 2026-04-24
maps: fix VLA stack overflow and sscanf type mismatches

CyberShadow:
Replace the per-iteration VLA `char filename[len]` with a single `char filename[PATH_MAX]` declared before the loop, preventing stack exhaustion when scanning processes with thousands of /proc/pid/maps entries (e.g. Wine).

Also fix the sscanf format string to match kernel field types: offset and inode are `unsigned long` (%lx/%lu), not `int` (%x/%u).

Fixes https://github.com/scanmem/scanmem/issues/452

Issue #452, opened 2025-11-05
Segmentation fault on expedition 33


hyunlee1o:
I am getting this error trying to open the process GameThread from expedition 33
```
sudo gameconqueror 88112
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en): No such file or directory

(GameConqueror.py:90217): Gtk-WARNING **: 22:42:57.938: Locale not supported by C library.
        Using the fallback 'C' locale.
/usr/share/gameconqueror/GameConqueror.py:1193: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://wiki.gnome.org/PyGObject/Threading
  GObject.threads_init()
/usr/share/gameconqueror/GameConqueror.py:1194: DeprecationWarning: Gdk.threads_init is deprecated
  Gdk.threads_init()
info: maps file located at /proc/88112/maps opened.
/usr/bin/gameconqueror: line 10: 90217 Segmentation fault         $PKEXEC $DATADIR/GameConqueror.py "$@"
```

a0kami:
hey @hyunlee1o, I had the exact same error, and it turns out for a very demanding game I turned zram on as a workaround
I tried by disabling zram (`sudo swapoff /dev/zram0`) and it now works properly, cheers

CyberShadow:
Stack overflow due to the /proc/pid/maps reader allocating paths on the stack in a loop. Fix: https://github.com/scanmem/scanmem/pull/455

Issue #450, opened 2025-06-09
GameConqueror fails for branch 0.17 on GTK 4.16.12-r1


koutoftimer:
I've configured custom prefix then followed install guide 

```
$ echo $PWD
/home/ruslan/opt/scanmem
$ ./configure --prefix="$PWD/dist" --enable-gui && make -j32
$ ./dist/bin/gameconqueror 

(GameConqueror.py:390008): Gtk-CRITICAL **: 13:07:27.986: gtk_clipboard_get_for_display: assertion 'display != NULL' failed
/home/ruslan/opt/scanmem/dist/share/gameconqueror/GameConqueror.py:1202: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://wiki.gnome.org/PyGObject/Threading
  GObject.threads_init()
/home/ruslan/opt/scanmem/dist/share/gameconqueror/GameConqueror.py:1203: DeprecationWarning: Gdk.threads_init is deprecated
  Gdk.threads_init()

(GameConqueror.py:390008): Gtk-CRITICAL **: 13:07:27.987: _gtk_css_lookup_resolve: assertion '(((__extension__ ({ GTypeInstance *__inst = (GTypeInstance*) ((provider)); GType __t = ((_gtk_style_provider_private_get_type ())); gboolean __r; if (!__inst) __r = (0); else if (__inst->g_class && __inst->g_class->g_type == __t) __r = (!(0)); else __r = g_type_check_instance_is_a (__inst, __t); __r; }))))' failed
/home/ruslan/opt/scanmem/dist/share/gameconqueror/GameConqueror.py:125: Warning: g_object_set_data_full: assertion 'G_IS_OBJECT (object)' failed
  self.builder.add_from_file(os.path.join(WORK_DIR, 'GameConqueror.ui'))

(GameConqueror.py:390008): Gtk-ERROR **: 13:07:27.987: Can't create a GtkStyleContext without a display connection
./dist/bin/gameconqueror: line 10: 390008 Trace/breakpoint trap   (core dumped) $PKEXEC $DATADIR/GameConqueror.py "$@"
```

AFAIK reporting issues that may not be present in master branch is not good but there is no similar issues at all so I'll leave it and verify that it is present in the master branch. 

master branch for commit `0375cc02b132c1678fcbeabdfee4cd81d353358f` has the same error message but line number is 126 instead of 125:

```
$ ./dist/bin/gameconqueror 

[...]

PR #457, opened 2026-08-01
scanmem 0.18.0: security hardening + backlog fixes

### LINKED ISSUE #449: .....*** buffer overflow detected ***: terminated
using CachyOS any time i try to scan for a number this happens

/usr/share/gameconqueror/GameConqueror.py:1193: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://pygobject.gnome.org/guide/threading.html#threads-faq
  GObject.threads_init()
/usr/share/gameconqueror/GameConqueror.py:1194: DeprecationWarning: Gdk.threads_init is deprecated
  Gdk.threads_init()
info: maps file located at /proc/29535/maps opened.
info: 185 suitable regions found.
info: maps file located at /proc/29535/maps opened.
info: 185 suitable regions found.
/usr/share/gameconqueror/GameConqueror.py:1111: DeprecationWarning: Gdk.threads_enter is deprecated
  Gdk.threads_enter()
/usr/share/gameconqueror/GameConqueror.py:1140: DeprecationWarning: Gdk.threads_leave is deprecated
  Gdk.threads_leave()
info: maps file located at /proc/29535/maps opened.
info: 185 suitable regions found.
.....*** buffer overflow detected ***: terminated
error: 
Killed by signal 6.

1rhino2:
## Summary

Bundles maintainer backlog fixes into one reviewable 0.18.0 prep branch (`miracle-update`).

### Security / stability
- Bounds checks in printable-string formatting (#449)
- Cap `/proc` maps pathname buffer; fix sscanf field types (#452, supersedes #455 maps half)
- Clamp ptrace poke length
- Process list via `/proc` instead of `ps` popen (#434)

~~Drop `system()` from shell handler; use `execvp`~~ reverted in 38795b8, see review discussion. It was not a security fix and it broke globs, pipes and variable expansion in `!` escapes.

### Compiler / portability
- `-Wclobbered` fixes in `handler__set` (#381)
- `sig_atomic_t` interrupt flag
- `volatile` scratch in `data_to_val_aux`
- Gtk 3 `require_version` before PyGI import (#398)
- Launch GameConqueror with `python3` (#438), and drop the py2 shims behind it
- Android cross compiler selection from `HOST` (#441), any API level. Logic tested against fake toolchain layouts, **not** verified on real hardware or a real NDK.

### GUI / i18n
- Memory viewer display type selector (#448, credit sqrt0x)
- Georgian translation (#426, credit NorwayFun)

### Release
- `configure.ac` bumped to 0.18.0
- `NEWS` entry for #397

## Build
Kali build with `-D_FORTIFY_SOURCE=3 -fstack-protector-strong` passes (`--disable-gui`).

## Not included (needs separate review)
#424 XOR, #403 monitoring, #389 undo, #348 stack offset, #346 read value, #312/#145 sm_reset, #450 GTK4.

Closes #449, #452, #438, #398, #441, #434 (partial via /proc list). Credits #448 #426 #381.

### DIFF
--- a/build_for_android.sh
+++ b/build_for_android.sh
@@ -149,18 +149,38 @@ if [ "$(uname -s)" = "Darwin" ]; then
   PATH=/usr/local/opt/gettext/bin:${PATH} # brew install gettext
 fi
 
-# NDK llvm toolchains need an explicit cross compiler or configure picks the host.
+# NDK r19+ names the compiler with an API level baked in (e.g.
+# aarch64-linux-android21-clang) and ships no unsuffixed symlink, so configure
+# quietly falls back to the host cc and builds for the wrong arch. Pick one
+# explicitly. Set CC yourself, or API_LEVEL, to override the search.
 if [ -z "${CC}" ]; then
   if command -v "${HOST}-clang" >/dev/null 2>&1; then
-    export CC="${HOST}-clang"
-  elif command -v "${HOST}21-clang" >/dev/null 2>&1; then
-    export CC="${HOST}21-clang"
-  elif command -v "${HOST}-gcc" >/dev/null 2>&1; then
-    export CC="${HOST}-gcc"
+    CC="${HOST}-clang"
+  elif [ -n "${API_LEVEL}" ] && command -v "${HOST}${API_LEVEL}-clang" >/dev/null 2>&1; then
+    CC="${HOST}${API_LEVEL}-clang"
   else
-    echo "Error: no cross compiler found for HOST=${HOST}" 1>&2
+    # no plain name and no API_LEVEL given, take the lowest API level on PATH
+    CC=$(IFS=:
+         for dir in $PATH; do
+           for cand in "${dir}/${HOST}"[0-9]*-clang; do
+             [ -x "${cand}" ] || continue
+             base=${cand##*/}
+             api=${base#${HOST}}
+             api=${api%-clang}
+             case "${api}" in *[!0-9]*) continue ;; esac
+             printf '%s %s\n' "${api}" "${base}"
+           done
+         done 2>/dev/null | sort -n | head -n 1 | cut -d' ' -f2)
+  fi
+  if [ -z "${CC}" ] && command -v "${HOST}-gcc" >/dev/null 2>&1; then
+    CC="${HOST}-gcc"
+  fi
+  if [ -z "${CC}" ]; then
+    echo "Error: no cross compiler found for HOST=${HOST}." 1>&2
+    echo "Set CC, or API_LEVEL if your NDK uses a versioned clang name." 1>&2
     exit 1
   fi
+  export CC
 fi
 export AR="${AR:-${HOST}-ar}"
 export RANLIB="${RANLIB:-${HOST}-ranlib}"
--- a/build_for_android.sh
+++ b/build_for_android.sh
@@ -148,6 +148,23 @@ fi
 if [ "$(uname -s)" = "Darwin" ]; then
   PATH=/usr/local/opt/gettext/bin:${PATH} # brew install gettext
 fi
+
+# NDK llvm toolchains need an explicit cross compiler or configure picks the host.
+if [ -z "${CC}" ]; then
+  if command -v "${HOST}-clang" >/dev/null 2>&1; then
+    export CC="${HOST}-clang"
+  elif command -v "${HOST}21-clang" >/dev/null 2>&1; then
+    export CC="${HOST}21-clang"
+  elif command -v "${HOST}-gcc" >/dev/null 2>&1; then
+    export CC="${HOST}-gcc"
+  else
+    echo "Error: no cross compiler found for HOST=${HOST}" 1>&2
+    exit 1
+  fi
+fi
+export AR="${AR:-${HOST}-ar}"
+export RANLIB="${RANLIB:-${HOST}-ranlib}"
+
 LIBS="-lncurses -lm" ./configure --host="${HOST}" --prefix="${SYSROOT}/usr" \
     --enable-static --disable-shared
 [ "$?" != "0" ] && exit 1
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,4 @@
-AC_INIT([scanmem],[0.18.0],[https://github.com/scanmem/scanmem])
+AC_INIT([scanmem],[0.18~dev],[https://github.com/scanmem/scanmem])
 AC_CONFIG_MACRO_DIR([m4])
 AM_INIT_AUTOMAKE([-Wall -Werror -Wno-extra-portability foreign])
 AC_CONFIG_SRCDIR([main.c])
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,4 @@
-AC_INIT([scanmem],[0.18~dev],[https://github.com/scanmem/scanmem])
+AC_INIT([scanmem],[0.18.0],[https://github.com/scanmem/scanmem])
 AC_CONFIG_MACRO_DIR([m4])
 AM_INIT_AUTOMAKE([-Wall -Werror -Wno-extra-portability foreign])
 AC_CONFIG_SRCDIR([main.c])
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
     Game Conqueror: a graphical game cheating tool, using scanmem as its backend
     
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -46,6 +46,7 @@ import misc
 
 import locale
 import gettext
+import pwd
 
 # In some locale, ',' is used in float numbers
 locale.setlocale(locale.LC_NUMERIC, 'C')
@@ -116,6 +117,38 @@ TYPESIZES = {'int8':1
             ,'float32':4
             ,'float64':8
             }
+
+def _read_proc_cmdline(pid):
+    cmdline_path = os.path.join('/proc', str(pid), 'cmdline')
+    try:
+        with open(cmdline_path, 'rb') as fh:
+            raw = fh.read()
+    except OSError:
+        raw = b''
+    if raw:
+        return raw.replace(b'\0', b' ').decode('utf-8', 'replace').strip()
+    exelink = os.path.join('/proc', str(pid), 'exe')
+    if os.path.exists(exelink):
+        return os.path.realpath(exelink)
+    return ''
+
+
+def _read_proc_username(pid):
+    status_path = os.path.join('/proc', str(pid), 'status')
+    try:
+        with open(status_path, 'r', encoding='utf-8', errors='replace') as fh:
+            for line in fh:
+                if line.startswith('Uid:'):
+                    parts = line.split()
+                    if len(parts) >= 2:
+                        try:
+                            return pwd.getpwuid(int(parts[1])).pw_name
+                        except (KeyError, ValueError):
+                            return parts[1]
+    except OSError:
+        pass
+    return ''
+
 
 class GameConqueror():
     def __init__(self):
@@ -916,18 +949,14 @@ class GameConqueror():
 
     def get_process_list(self):
         plist = []
-        for proc in os.popen('ps -wweo pid=,user:16=,command= --sort=-pid').readlines():
-            try:
-                (pid, user, pname) = [tok.strip() for tok in proc.split(None, 2)]
-            # process name may be empty, but not the name of the executable
-            except (ValueError):
-                (pid, user) = [tok.strip() for tok in proc.split(None, 1)]
-                exelink = os.path.join("/proc", pid, "exe")
-                if os.path.exists(exelink):
-                    pname = os.path.realpath(exelink)
-                else:
-                    pname = ''
-            plist.append((int(pid), user, pname))
+        for entry in os.listdir('/proc'):
+            if not entry.isdigit():
+                continue
+            pid = int(entry)
+            user = _read_proc_username(pid)
+            pname = _read_proc_cmdline(pid)
+            plist.append((pid, user, pname))
+        plist.sort(key=lambda item: item[0], reverse=True)
         return plist
 
     def select_process(self, pid, process_name):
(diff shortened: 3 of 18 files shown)


1rhino2:
Went through this end to end before asking anyone to merge it.

Reviewed all 18 changed files line by line (handlers.c, maps.c, ptrace.c, targetmem.c, common.c, interrupt.c, the GUI python files, build_for_android.sh). Nothing surprising, matches what the summary describes.

Build:
- `./autogen.sh && ./configure --disable-gui CFLAGS="-D_FORTIFY_SOURCE=3 -fstack-protector-strong -Wall -Wextra"` then `make` - clean, one pre-existing `-Wuse-after-free` warning in `targetmem.h:allocate_enough_to_reach` that's on `main` too, not touched by this branch (separate issue, will file).
- `make check` - passes, including the 257-byte forced-fullscan bytearray/string cases in `sm_test.sh`.
- Rebuilt with `-fsanitize=address,undefined -g -O1`, reran `make check` - clean.

Manual smoke against `memfake`:
- `!` shell handler (fork/execvp replacing `system()`): runs commands, correct exit-status handling, and correctly errors instead of silently falling through to a shell when the binary doesn't exist.
- `list` with a 300-byte bytearray/string scan (exercises the realloc + `append_suffix` path that replaced the old `strcat`) - no truncation bugs, no ASan hits.
- `write int8` / `write string` on a live match address under ASan - clean.

Only thing I'd flag as follow-up, not a blocker: that pre-existing UAF warning in `targetmem.h`. Opening that separately since it's unrelated to this branch's diff.

gui/GameConqueror.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
12345ieee, line 1: Man, I spent a good bunch of time trying to make the script both py2 and py3 compatible.

Good call here, it's time

gui/misc.py
```diff
@@ -21,6 +21,8 @@
 
 import sys
 
+import gi
+gi.require_version('Gtk', '3.0')
 from gi.repository import Gtk
 
 PY3K = sys.version_info >= (3, 0)
```
12345ieee: Should be able to drop this and assume py3 everywhere.

build_for_android.sh
```diff
@@ -148,6 +148,23 @@ fi
 if [ "$(uname -s)" = "Darwin" ]; then
   PATH=/usr/local/opt/gettext/bin:${PATH} # brew install gettext
 fi
+
+# NDK llvm toolchains need an explicit cross compiler or configure picks the host.
```
12345ieee: Has this actually been tested? I don't own a linux phone, never could.

handlers.c
```diff
@@ -1127,29 +1166,33 @@ bool handler__shell(globals_t * vars, char **argv, unsigned argc)
...
+    child = fork();
+    if (child == -1) {
+        show_error("fork() failed, command was not executed.\n");
+        return false;
+    }
 
-    /* allocate space */
-    command = calloca(len, 1);
+    if (child == 0) {
+        execvp(argv[1], argv + 1);
```
12345ieee: Doing it this way drops the shell, so it's technically different.

Are you doing this for security reasons?

12345ieee (commented):
Did a cursory review, I apparently recall GC much more I do scanmem.

handlers.c
1rhino2: yeah you're right, reverted in 38795b8.

it isn't a security fix. it's the user's own typed command in their own session, there's no injection boundary there. i saw `system()` and pattern matched on it.

and it broke more than i realised:

```
> ! echo $HOME
$HOME
> ! ls *.ac
ls: cannot access '*.ac': No such file or directory
> ! echo one | tr a-z A-Z
one | tr a-z A-Z
```

globs, pipes, `$vars`, all dead, on the shell escape command. the `XXX: handle !ls style escapes` comment sitting right above it was the hint i ignored.

old buffer math was fine too fwiw, `len` works out to exactly `sum + argc` which is what the concat writes. so nothing was being fixed either way. handler matches main again now.

build_for_android.sh
1rhino2: no. no NDK and no android device here, nothing's been built for a phone. should've said that plainly instead of leaving it in a list that reads as tested.

did pull the block out and run it against fake toolchains on PATH:

- only `21-clang` present, picks it
- only `29-clang`, picks it
- 21 + 24 + 29, picks 21
- `API_LEVEL=24`, picks 24
- plain `-clang` present, wins over versioned
- nothing there, errors instead of quietly using the host cc

that last one is the actual #441 bug.

doing that also caught my own screwup, i'd hardcoded `21` so an NDK shipping only 24 or 29 still fell through. fixed in a91316f.

so the logic works, a real cross build doesn't exist. if you'd rather not carry something untested on hardware i'll pull this commit out and leave #441 open.

gui/misc.py
1rhino2: done in 7f943e5. `PY3K` gone, decode/encode/str2bytes collapse to their py3 bodies, both `long` checks go, and `import sys` was only there for the version check. GameConqueror.py had two more, the guint64 conversion is unconditional now.

ran the rewritten ones rather than trust them: eval_operand on ints/floats/`1<<40`, check_int taking a good int32 and rejecting both an overflow and a float, decode/encode round trip. same behaviour as the py3 path had.

gui/GameConqueror.py
1rhino2: ha, sorry for retiring it. took the rest of the py2 stuff out too, see the misc.py thread.

1rhino2:
pushed three.

38795b8 puts the shell handler back to `system()`, that one was just wrong, details in the thread. 7f943e5 kills the py2 shims. a91316f fixes the android compiler search, which wasn't only about HOST, i'd pinned API 21.

reran everything after the revert, `make check` passes, asan/ubsan build clean, and `$HOME` / globs / pipes / `&&` all work again.

so what's actually left under the security heading is the bounds work, which is where the reported crashes were: #449 formatters writing past `buf_length`, #452 the VLA in `sm_readmaps` plus the sscanf types, the memcpy clamps in `handler__watch` and `sm_setaddr`, and the strcpy/strncpy pair. rest of the PR is portability, gui, and the three contributor PRs folded in.

two i'm not touching here, can file separately: the `-Wuse-after-free` in `targetmem.h` is pre-existing on main, and `hexview.py:114` has an unescaped `\]` that throws a SyntaxWarning on py3, from a 2010 commit.

CI still needs your approval to run so none of this has been through it.

1rhino2:
reran everything on the current head of this branch so you have real numbers instead of my word for it.

builds clean on gcc 15.3 (autogen, configure, make -j4, all exit 0). make check passes 1/1.

end to end smoke against a live process, not just a compile: started a target holding a known int, scanned for it, got matches, wrote a new value, target read the new value back and exited on it. so maps parsing and the ptrace write path both actually work.

also specifically retested the shell handler, since 38795b8 reverts my earlier execvp change back to system(). pipes, $var expansion and globs all still work (`shell echo a b c | tr " " "_"`, `shell echo $HOME`, `shell ls *.c`). that revert was the right call and the mistake was mine, the execvp version would have silently killed all three on the one command whose entire job is running a shell. there is a `XXX: handle !ls style escapes` note sitting right above that function that i should have read the first time.

one thing i hit that is not from this pr so it should not block it: targetmem.h:169 throws -Wuse-after-free on gcc 12+, the pointer arithmetic against the pre realloc pointer used to fix up interior swath pointers. same lines on main (141 and 169), so it predates this branch. left it alone to keep this diff scoped, happy to do it as its own pr after.

@12345ieee no rush on the review. sriemer sent the owner bit but i would rather you actually look at a 1400 line diff than have me merge my own on day one. the three prs it supersedes (#381, #448, #426) stay open until this lands.

1rhino2:
ci is green now: https://github.com/scanmem/scanmem/actions/runs/32197701283

worth flagging why it had never run. every previous run on this pr sat at `action_required` with 0s, so the workflow was gated behind approval the whole time and nobody had clicked it. i approved this one now that i have the bit. that build is not a soft one either, it compiles with `-fsanitize=address,undefined` and runs `sudo make check`, so asan and ubsan are both clean on top of the local run i posted above.

not touching the approval gate setting itself. it looks like a misconfiguration at first glance but it is the right default here, 01-coverity-scan.yml runs against a scan token and auto running fork prs would hand that to anyone who opens one. approving runs by hand is the correct tradeoff, it is just maintainer work rather than something to configure away.

1rhino2:
correcting myself on the ci gate reasoning above. i said auto running fork prs would hand out the coverity scan token. that part is wrong, 01-coverity-scan.yml only triggers on branches matching `**coverity**` or manual dispatch, so a fork pr never reaches it and never sees that secret.

the gate should still stay, just for the ordinary reason rather than the one i gave: 01-build.yml runs whatever is in the pr and ends with `sudo make install`, so approving a run means agreeing to execute a strangers code on the runner. that is worth a human glance for drive by prs regardless of secrets.

conclusion is the same, the reasoning i posted was not.

1rhino2:
status, and then i will stop pinging this.

main moved under this today, #458 (realloc UB in targetmem) and #459 (clearer attach errors, closes #392 and #393) both landed. i checked this branch against the new main and it still merges clean, no conflicts with either, so nothing here needs a rebase. main also builds with zero warnings now, this branch included.

@12345ieee if you want to read it, it is ready and it is yours, take the time. if you would rather not, that is completely fine too, you already said you are not keeping up and nobody is owed unpaid weekend review.

so that this does not just sit: if i have not heard from you by sunday the 23rd i will merge it and carry it. not to rush you, it is that eight other prs are stacked behind this one touching handlers.c and they cannot be rebased until it lands. if you want longer, say so and it waits, a word is enough.

the parts worth your eyes if you do look, rather than all 16 commits: feacf33 (maps pathname cap and the sscanf types) and f9000cf (bounds checks). the rest is gui, packaging and translations.

12345ieee:
I will hopefully have a look in the weekend.
Not just at this, but at everything that went in in the meantime.

The project needed it, just allow me some time to do due diligence.

1rhino2:
take the weekend, the sunday thing is off. you said you want to do due diligence on everything that landed, that is the right call and i would rather this get read properly than land on a date i picked.

waiting does not cost anything either, i had that wrong. i trial merged the stacked work against this branch today and it is 3 conflict hunks across 2 files, all of them the same bug fixed twice, resolved in about five minutes. the rebase pileup i used as the reason for a deadline is not real.

one correction before you read anything, in case you saw me say otherwise elsewhere. the handler__set change in here, `unsigned block, seconds` and `bool cont` from #381, is fine as it stands. i claimed it silences -Wclobbered without fixing the underlying indeterminacy. that was wrong: the interrupt path only calls sm_detach, ENDINTERRUPTABLE and returns, so neither variable is ever read after the longjmp and there is nothing to observe. no action needed, i just do not want you spending review time on a problem i invented.

separately, main has no working CI right now, the last two runs there hung for six hours and were cancelled. that is unrelated to this branch and i put a two file fix on `ci-unhang` off main so it can go in on its own.

12345ieee:
A PR this big and varied is a bit hard to review, let's do it step by step.

The GC changes seem fine and work properly, there is
```
scanmem/gui/GameConqueror.py:1236: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://pygobject.gnome.org/guide/threading.html#threads-faq
  GObject.threads_init()
scanmem/gui/GameConqueror.py:1237: DeprecationWarning: Gdk.threads_init is deprecated
  Gdk.threads_init()
```
to address in a nxt changeset.

configure.ac
```diff
@@ -1,4 +1,4 @@
-AC_INIT([scanmem],[0.18~dev],[https://github.com/scanmem/scanmem])
```
12345ieee, line 1: Probably better to change this when you've merged everything you have and want to cut a 0.18.

1rhino2:
configure.ac reverted to 0.18~dev in 3e9b38d. it goes to 0.18.0 when you're ready to cut the release and have merged what you want first. i also dropped the fixed date on the NEWS section so it reads "0.18.0 (unreleased)" rather than claiming a ship date. the content stays as prepared notes, say if you'd rather i pull the whole section until release.

the threads_init deprecation you spotted at 1236/1237 is handled in #466, already up and green, that's the next changeset you're describing. it removes the two no-op threads_init calls, but the real work is the threading under them. those calls propped up the old Gdk.threads_enter/leave model, and the scan worker was touching GTK widgets off the main thread under that lock. the lock is gone from modern GTK, so that races the main loop and segfaults, which is the crash behind #427, #413, #435 and #450. #466 moves the worker off the GTK path with GLib.idle_add, and then threads_init has nothing left to prop up, so it goes.

i kept that out of this PR on purpose, a behaviour change to the threading has no business buried in the 0.18 prep diff. numbers if you want them: a 40 scan stress loop against memfake segfaults 4/4 runs on current main and runs 3/3 clean with #466 on top.

so nothing to add to this branch for it, it lives next door.

12345ieee:
Looked at the rest, everything looks reasonable and some cursory manual testing went ok.

At this point LLMs are better than me at finding buffer overruns, so if some are left they'll get them.

You can push this in, use the commit reshaping technique you prefer (squash, as-is, partial commit fusions, ...), just please do not use merge commits.

Point me to the next thing in line I should look at, and consider adding a `.clang-format` in another PR.
Tooling got a lot better in the last decade, my god it's been a decade.

1rhino2:
merged as 2341d8e, squashed to one commit, no merge commit. built and make check clean on main after.

next in line: #466, the GUI threading one. small, one file, and it is the fix under the threads_init warning you flagged here. the scan worker was touching GTK off the main thread under the old Gdk.threads_enter/leave, which is the segfault in #427, #413, #435 and #450. it moves that onto the main loop with idle_add. the thing worth your eye is that it is a real behaviour change to the threading, not a cosmetic cleanup, so the before/after is what to check: a 40 scan stress loop crashes 4/4 on current main and runs 3/3 clean with it. after that, #465 carries the peekbuf out of bounds write, which is the buffer overrun kind you said you would rather the tools catch.

on .clang-format: will do, as its own PR so the reformat churn does not ride on anything else. i will match what the tree already does, four space, K&R-ish braces, rather than impose a new style, and keep it to the config plus a one time format so it stays reviewable.

no rush on the reviews, land them when you have time.

PR #463, opened 2026-08-22
ptrace: fix the out of bounds write in the peekbuf head shift

1rhino2:
Found the cause of #449, the `*** buffer overflow detected ***` on scan.

sm_peekdata mirrors target memory in a 67584 byte cache. On a partial hit it
works out how much more it needs, rounds up to a whole chunk, and if that will
not fit it shifts the frame down. The shift only moves whole chunks.

peekbuf.size stays a multiple of the chunk size right up until a read comes
back short, which happens when the read runs into an unmapped page. That
remainder then sits in size, the shift cannot move it, and the room it frees
is short by exactly that much. The read loop runs anyway and writes past the
end of the array. Worst case with a full length request is 2045 bytes.

You need the page missing on one read and present on the next, so a fully
stopped target never gets there. Attaching stops only the main thread though,
so a game's other threads keep allocating through the scan. That fits the
reports: fine on one box, aborts on another, and _FORTIFY_SOURCE builds turn
it into the abort in the issue, since glibc checks the pread destination.

Asan on current main, running the test in this PR:

```
ERROR: AddressSanitizer: global-buffer-overflow
WRITE of size 2048 ... in sm_peekdata
0 bytes after global variable 'peekbuf' defined in 'ptrace.c' of size 67608
```

Fix checks the fit again after the shift and drops the cache if it still does
not fit. A fresh read is bounded by length, a uint16_t, so it always fits.

test/peekbuf_test.c reproduces it with a child that maps a run ending part way
through a chunk, then remaps the hole. Fails on main with and without
hardening, passes with the fix.

Match counts unchanged over six interleaved runs of both builds against one
memfake across int8, int, float, number, bytearray, string and snapshot.

### DIFF
--- a/ptrace.c
+++ b/ptrace.c
@@ -310,6 +310,18 @@ extern inline bool sm_peekdata(const void *addr, uint16_t length, const mem64_t 
 
             peekbuf.size -= shift_size;
             peekbuf.base += shift_size;
+
+            /* the shift only frees whole chunks, so it is short by whatever
+             * partial chunk a truncated read left in `size`. rather than
+             * reason about how short, check again and drop the cache when it
+             * still does not fit. `length` is a uint16_t so a fresh read is
+             * always MAX_PEEKBUF_SIZE at worst. */
+            if (peekbuf.size + missing_bytes > MAX_PEEKBUF_SIZE)
+            {
+                peekbuf.size = 0;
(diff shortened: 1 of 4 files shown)


1rhino2:

Issue #322, opened 2018-08-15
Feature Request: Allow restricting searches to around an address on the cheat list.


BluecoatEngineer:
Let's say I'm in an arbitrary game, searching for some character stat like health.  I find it, lock it, and go on to try to search for ammo, mana, etc.  What's the likelihood that the representation of game state doesn't look something like:

    struct {
        int hit_points;
        int ammo;
        int and_then;
        int and_Then;
        int and_THEN;
        ...
    } game_state;

If we assume that to be the case, most of the other values of interest for cheating are likely to be near the first result.  It'd be a cool feature to be able to restrict searches to some +/- bytes from an entry in the cheat list and get nearly instantaneous results.

sriemer:
You could use the hex editor of GameConqueror or the `dump` command of scanmem.
Do you know that the locking/freezing of values in scanmem/GC only works once if it is not the unlikely case of all values in non-PIE executable?

For a real game trainer which bypasses ASLR/PIC/PIE and has dynamic memory support, look at [ugtrain](https://github.com/ugtrain/ugtrain). It dumps objects in dynamic memory and in pointer memory. This way two states can be compared and you can find the other interesting values within the objects.

BluecoatEngineer:
Huh, I didn't realize that GC had a memory editor!  I still think there's use for being able to tell it to search "around" the area, especially for progress bars and floats, but being able to explore around the address covers most of it.

ugtrain looks pretty cool, I'll check it out!

sriemer:
Okay, thanks. Marked as GUI enhancement.

mendelmunkis:
On the subject it may be good idea to have a user selectable search scope

sriemer:
This is selectable by `scanmem` command `dregion` on region level. E.g. `dregion !0,1` is often used to limit search results for pointers to the executable .data and .bss regions. Professionals don't need a GUI. The GUI can't cover all CLI features.

mendelmunkis:
I'm sorry I wasn't clear. I was referring to the ability to select an arbitrary address range to search. I also didn't mean to imply that it should necessarily be a GUI feature, I just thought that this discussion on limiting search scope was a good place to bring it up.

PR #464, opened 2026-08-22
handlers: add `range` to crop the scan to an address window

### LINKED ISSUE #429: Feature request, specfying the extact range of addresses to scan
At the moment GC only permits a slider for small/medium/large range of addresses to search through, I happen to know rough start/end range I need to scan is just barely above the 32bit range of 0xFFFF'FFFF in the 0x5'0000'0000 to 0x6'FFFF'FFFF range which currently forces me to use the large scan which takes way longer than what it would if I could just limit the range to what I need it to scan specifically, you could swap the slider for a drop-down that has a "Custom" option to enable 2 textboxes for manually entering the range while the original 3 options would disable them and override what they display.

1rhino2:
Closes #429, and covers #322 as well.

`dregion` can only drop whole regions, so there is no way to say "it is somewhere between 0x500000000 and 0x700000000, only look there". On a large process that is the difference between reading a few hundred MB and reading the lot.

`range <start> <end>` intersects every region with the window. Regions fully outside are dropped, regions crossing an edge are cut down to the part inside, and matches that fall outside go with them so the match list stays consistent with the regions it came from. Both addresses hex, start inclusive, end exclusive, and `reset` puts the full list back.

Offsets in `list` come from load_addr rather than the region start, so cropping does not shift what you see.

For the GUI side of #429 this is the primitive the two text boxes would call.

test/range_tests.sh checks the part that is easy to get wrong. Splitting a 64k window in half has to account for every match the whole window finds:

```
window 7f3817388000-7f3817398000: whole=65533 first=32765 second=32768
```

32765 + 32768 = 65533, the three missing being the ints that straddle the split point. It also checks that cropping an existing match list leaves the same count as scanning the cropped window from scratch, that an unmapped window warns and finds nothing, that reset restores the regions, and that backwards, non hex and missing arguments are refused.

Both halves fail that suite when broken on purpose, the crop and the match deletion separately, so it is not passing vacuously. It skips rather than fails when it cannot ptrace, since a plain `make check` is not root and CI runs it under sudo.

valgrind over scan, crop, rescan, crop to nothing, reset: no errors, no definite leaks. Clean under -Wextra. Man page entry added.

### DIFF
--- a/handlers.c
+++ b/handlers.c
@@ -668,6 +668,101 @@ bool handler__dregion(globals_t *vars, char **argv, unsigned argc)
     return true;
 }
 
+/* drop matches that fall in [from, to), if there are any to drop */
+static void forget_matches(globals_t *vars, unsigned long from, unsigned long to)
+{
+    if (vars->num_matches == 0 || from >= to)
+        return;
+
+    vars->matches = delete_in_address_range(vars->matches, &vars->num_matches,
+                                            (void *)from, (void *)to);
+    if (vars->matches == NULL)
+        show_error("memory allocation error while deleting matches\n");
+}
+
+bool handler__range(globals_t * vars, char **argv, unsigned argc)
+{
+    unsigned long lo, hi;
+    char *endptr;
+    element_t *np, *pp = NULL;
+    size_t kept = 0, cropped = 0, dropped = 0;
+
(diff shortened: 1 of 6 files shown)


1rhino2:

PR #465, opened 2026-08-22
CI unhang, the #449 out of bounds write, and the range command

1rhino2:
#460, #463 and #464 in one branch so it is one merge instead of three. Nothing new here, same commits, plus the one line in test/Makefile.am where #463 and #464 both added a test.

What is in it:

CI first, because main's CI is currently dead. A step with no timeout burned six hours before someone cancelled it, and `apt upgrade` in the setup action is what wedges. Timeout capped at 30 minutes, upgrade dropped, apt retries added.

The #449 crash. Real out of bounds write in sm_peekdata, not a packaging problem. A short read leaves a partial chunk in peekbuf.size, the head shift only moves whole chunks, so the room it frees is short by that remainder and the following read is set up to write past the end of the array. Up to 2045 bytes on a full length request. Asan on main, running the test that comes with this:

```
ERROR: AddressSanitizer: global-buffer-overflow
WRITE of size 2048 ... in sm_peekdata
0 bytes after global variable 'peekbuf' defined in 'ptrace.c' of size 67608
```

Only reachable when the mapping is absent on one read and back on the next, so a stopped target never sees it. Attaching stops the main thread only, a game's other threads carry on.

`range <start> <end>` for #429 and #322, crops the regions list to an address window so a scan does not have to read the whole address space. Splitting a 64k window in half accounts for every match the whole window finds, 32765 + 32768 = 65533, the three missing being the ints straddling the split.

Both new tests fail when the thing they cover is broken on purpose, so they are not vacuous. valgrind clean over the range paths. `sudo make check` exits 0 here, po/ check included.

### DIFF
--- a/.github/actions/linux-setup/action.yml
+++ b/.github/actions/linux-setup/action.yml
@@ -4,15 +4,17 @@ description: Set up the Linux VM
 runs:
   using: 'composite'
   steps:
-    # Update packages
-    - run: sudo apt --assume-yes update
+    # Update packages. Retries because a mirror hiccup here used to wedge the
+    # whole run, and noninteractive so nothing can sit waiting on a prompt.
+    - run: sudo apt-get -o Acquire::Retries=3 --assume-yes update
       shell: bash
-    # Upgrade packages
-    - run: sudo apt --assume-yes upgrade
-      shell: bash
+      env:
+        DEBIAN_FRONTEND: noninteractive
     # Install required packages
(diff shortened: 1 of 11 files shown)


12345ieee:
Sorry am late, please allow me a couple extra days, if you don't hear from me this week you can start merging some yourself.

1rhino2:
> Sorry am late, please allow me a couple extra days, if you don't hear from me this week you can start merging some yourself.

All good no worries

Issue #435, opened 2023-07-30
GC errors on clipboard


Linrox:
Trying to run game conqueror errors 

`Traceback (most recent call last):
  File "/usr/share/gameconqueror/GameConqueror.py", line 54, in <module>
    CLIPBOARD = Gtk.Clipboard.get(Gdk.SELECTION_CLIPBOARD)
  File "/usr/lib64/python2.7/site-packages/gi/module.py", line 132, in __getattr__
    self.__name__, name))
AttributeError: 'gi.repository.Gdk' object has no attribute 'SELECTION_CLIPBOARD'
`
Not sure if GC will work for leap 15.4. Scanmem works without issue.

1rhino2:
that traceback is python 2, the path gives it away:

```
File "/usr/lib64/python2.7/site-packages/gi/module.py", line 132, in __getattr__
AttributeError: 'gi.repository.Gdk' object has no attribute 'SELECTION_CLIPBOARD'
```

gameconqueror shipped a `#!/usr/bin/env python` shebang, so on a box where `python` still pointed at 2.7 it got launched under py2 and the gi bindings there do not expose that constant. the clipboard line is not actually wrong.

#457 changes the shebang to python3 and makes the launcher invoke python3 explicitly rather than trusting whatever `python` resolves to, which should sort this. same change fixes #438, which is the other side of the same coin, boxes with no `python` at all.

leaving this open rather than closing it on a guess, i do not have a leap 15.4 box to confirm on. if you are still around, retesting once 0.18.0 is tagged would be useful. separately, the "scanmem keeps asking for PID without progressing" at the end of your report is unrelated to the clipboard crash, that is usually the attach failing, and #459 just landed so the error will now tell you which of the two causes it actually is.

Issue #434, opened 2023-07-17
ps issue


Melechtna:
GameConqueror.py:104297): dbind-WARNING **: 14:32:09.209: AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: The name org.a11y.Bus was not provided by any .service files
/nix/store/08kl4dq5znqgdzv1pm73i1v2z0wqpa4w-scanmem-0.18-dev/share/gameconqueror/GameConqueror.py:1193: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://wiki.gnome.org/PyGObject/Threading
  GObject.threads_init()
/nix/store/08kl4dq5znqgdzv1pm73i1v2z0wqpa4w-scanmem-0.18-dev/share/gameconqueror/GameConqueror.py:1194: DeprecationWarning: Gdk.threads_init is deprecated
  Gdk.threads_init()
Traceback (most recent call last):
  File "/nix/store/08kl4dq5znqgdzv1pm73i1v2z0wqpa4w-scanmem-0.18-dev/share/gameconqueror/GameConqueror.py", line 529, in SelectProcess_Button_clicked_cb
    pl = self.get_process_list()
  File "/nix/store/08kl4dq5znqgdzv1pm73i1v2z0wqpa4w-scanmem-0.18-dev/share/gameconqueror/GameConqueror.py", line 913, in get_process_list
    (pid, user, pname) = [tok.strip() for tok in proc.split(None, 2)]
ValueError: not enough values to unpack (expected 3, got 0)
Traceback (most recent call last):
  File "/nix/store/08kl4dq5znqgdzv1pm73i1v2z0wqpa4w-scanmem-0.18-dev/share/gameconqueror/GameConqueror.py", line 529, in SelectProcess_Button_clicked_cb
    pl = self.get_process_list()
  File "/nix/store/08kl4dq5znqgdzv1pm73i1v2z0wqpa4w-scanmem-0.18-dev/share/gameconqueror/GameConqueror.py", line 913, in get_process_list
    (pid, user, pname) = [tok.strip() for tok in proc.split(None, 2)]
ValueError: not enough values to unpack (expected 3, got 0)

Seems to be an issue with the assumptions made in the ps parser.

Issue #427, opened 2023-03-13
Gameconqueror "Authorization required, but no authorization protocol specified"


DNDEBUG:
```
Authorization required, but no authorization protocol specified

Authorization required, but no authorization protocol specified


(GameConqueror.py:31788): Gtk-CRITICAL **: 21:49:34.693: gtk_clipboard_get_for_display: assertion 'display != NULL' failed
/usr/share/gameconqueror/GameConqueror.py:1193: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://wiki.gnome.org/PyGObject/Threading
  GObject.threads_init()
/usr/share/gameconqueror/GameConqueror.py:1194: DeprecationWarning: Gdk.threads_init is deprecated
  Gdk.threads_init()

(GameConqueror.py:31788): Gtk-CRITICAL **: 21:49:34.698: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed

(GameConqueror.py:31788): Gtk-CRITICAL **: 21:49:34.698: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed

(GameConqueror.py:31788): Gtk-CRITICAL **: 21:49:34.698: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed
/usr/bin/gameconqueror, linha 10: 31788 Falha de segmentação  $PKEXEC $DATADIR/GameConqueror.py "$@"
````

aaruni96:
I am running into the same problem (running on Hyprland)

Seme4eg:
same here

sriemer:
Seems like `pkexec` changes. Might have been fixed in the Debian packages already.

aaruni96:
As a workaround, if you execute `xhost +SI:localuser:root` before trying to run gameconquerer (or any other GUI app which requires root permissions), it works on wayland. I don't know what it means, I hope its helpful.

Seme4eg:
`xhost +local: &; gameconqueror` - that's what also worked for me. Needed to install xorg-xhost first tho.

nerozero:
The issue is still here

Arch, Wayland, Hyprland

sriemer:
[...]

Issue #413, opened 2022-07-28
Fails to launch from root user in SteamOS


Silfurfjoss:
When attempting to launch as root user (I was testing to see if I could bypass requiring password), it fails with a few critical errors, pasted below. This was attempting to launch from konsole. If I launch it as normal user and supply the password it launches as expected from konsole. 

(GameConqueror.py:4657): Gtk-CRITICAL **: 17:04:49.394: gtk_clipboard_get_for_display: assertion 'display != NULL' failed
/usr/share/gameconqueror/GameConqueror.py:1193: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://wiki.gnome.org/PyGObject/Threading
  GObject.threads_init()
/usr/share/gameconqueror/GameConqueror.py:1194: DeprecationWarning: Gdk.threads_init is deprecated
  Gdk.threads_init()

(GameConqueror.py:4657): Gtk-CRITICAL **: 17:04:49.401: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed

(GameConqueror.py:4657): Gtk-CRITICAL **: 17:04:49.401: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed

(GameConqueror.py:4657): Gtk-CRITICAL **: 17:04:49.401: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed
/usr/bin/gameconqueror: line 10:  4657 Segmentation fault      (core dumped) $PKEXEC $DATADIR/GameConqueror.py "$@"
(139)(B+)(root@steamdeck ~)#

1rhino2:
Same cause as #427, the display cannot be opened so the first clipboard call segfaults. #462 stops the crash and prints what to do instead, and points XAUTHORITY at your cookie before pkexec so root can actually find it. I do not have a Deck, so if you get a chance to try it after that lands I would like to know whether it comes up or just gives you the message.

PR #466, opened 2026-08-22
gui: drive GTK from the main thread instead of the scan worker

1rhino2:
Fixes the segfault-on-scan cluster: #427, #413, #435, #450.

GameConqueror still used the PyGTK-era `Gdk.threads_enter()` / `Gdk.threads_leave()` to let the scan worker thread touch widgets. That global lock is gone from GTK, so on a modern GTK3 the calls serialize nothing and the worker mutating widgets races the main loop. That is the `Gtk-CRITICAL ... assertion 'GTK_IS_...' failed` spew and then the segfault in nearly every one of those reports.

Three enter/leave sites, one real thread:

- `progress_watcher` and `data_worker` already run on the main loop via `GLib.timeout_add`, so their locking was dead code. Removed.
- `scan_thread_func` is the actual worker. It now does only the blocking `backend.send_command`, then marshals the widget updates to the main loop with `GLib.idle_add(self._on_scan_finished)`. Nothing touches GTK off-thread anymore.

`command_lock` is released before the idle callback so `update_scan_result` re-taking it runs on the main thread without deadlocking, and `source_remove` of the progress watcher moves to the main thread too.

Also dropped `GObject.threads_init` / `Gdk.threads_init` in main. No-ops since PyGObject 3.11, they only warned every launch (#434), and nothing needs them once the worker is off the GTK path. This also unblocks GTK4, where `threads_enter/leave` do not exist.

Reproduced it: a headless driver runs 40 scans in a row against a memfake target with a locked cheat row so `data_worker` polls concurrently.

```
old code (origin/main): 4/4 runs died  (SIGSEGV x3, SIGABRT x1)
this branch:            3/3 runs clean, 0 Gtk-CRITICAL, 40/40 scans, 202 matches each
```

No behaviour change to scanning, locking, or the cheat list. Branched off main so it is independent of the backlog PRs; happy to fold it into #462 if you would rather it ride with the other GUI fixes.

### DIFF
--- a/gui/GameConqueror.py
+++ b/gui/GameConqueror.py
@@ -595,10 +595,9 @@ class GameConqueror():
     def scanresult_delete_selected_matches(self, menuitem, data=None):
         (model, pathlist) = self.scanresult_tv.get_selection().get_selected_rows()
         match_id_list = ','.join(str(model.get_value(model.get_iter(path), 6)) for path in pathlist)
-        self.command_lock.acquire()
-        self.backend.send_command('delete {}'.format(match_id_list))
-        self.update_scan_result()
-        self.command_lock.release()
+        with self.command_lock:
+            self.backend.send_command('delete {}'.format(match_id_list))
+            self.update_scan_result()
 
     def scanresult_popup_cb(self, menuitem, data=None):
         (model, pathlist) = self.scanresult_tv.get_selection().get_selected_rows()
@@ -924,10 +923,9 @@ class GameConqueror():
             self.process_label.set_text('%d - %s' % (pid, process_name))
             self.process_label.set_property('tooltip-text', process_name)
 
-        self.command_lock.acquire()
-        self.backend.send_command('pid %d' % (pid,))
-        self.reset_scan()
-        self.command_lock.release()
+        with self.command_lock:
+            self.backend.send_command('pid %d' % (pid,))
+            self.reset_scan()
 
         # unlock all entries in cheat list
         for i in range(len(self.cheatlist_liststore)):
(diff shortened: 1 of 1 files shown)


12345ieee:
While the comments about previous code state help my review, I assume you'll cut them by the merge as they don't help a future reader.

Code looks sensible, crashes disappeared, this will close a ton of issues.

Small nit, if you move to `with self.command_lock:`, which is a good idea, change the other manual acquire release as well.

On to the next, what's up?

12345ieee:
Do you want me to make the change and merge?

1rhino2:
pushed both. trimmed the comments down to the forward-looking ones and dropped the ones that only narrated the old code, and moved the rest of the manual acquire/release over to with self.command_lock: so it matches the worker.

one exception: data_worker uses a non-blocking acquire(0) to skip a tick if a scan is holding the lock, and with cant express skip-if-busy, so i kept the conditional but put the held region under try/finally. same effect, and it now releases if the body throws.

reran the headless stress loop after the change, 3/3 clean, 40 scans each, no Gtk-CRITICAL.

so just the merge left. go ahead whenever youre happy, or i can once CI is green, whichever you prefer. figure the next thing after this is lining up 0.18 with these crash fixes in it.

12345ieee:
Nice, shipping.

Before 0.18 I'd merge more stuff, what PR would you like me to see next?

1rhino2:
#465 next if you're up for it. it's the #449 out-of-bounds write, plus the CI unhang and the range command. all C side so it doesn't touch anything #466 did, and it's green and mergeable right now. that's the one i'd most want in 0.18 since the OOB write is a real fix.

#461 (the scan perf rework) is the bigger one and still a draft, i'd leave that for after 0.18 rather than rush it in.

